PR #50, opened 2014-06-28
allow map[string]interface{} to respect concrete "schema" types

mechmind:
Hello. I want to decode into concrete types, stored in `map[string]interface{}`, so I implemented a "schema" values - when `rv` is `map[string]interface{}` and for given key in config we have a value in map, instantiate new value of that type and unify it, otherwise instantiate and unify the new value of map elem type. The keys in `rv`, not found in config are dropped.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -278,7 +278,7 @@ func (md *MetaData) unifyMap(mapping interface{}, rv reflect.Value) error {
 		rvkey.SetString(k)
 
 		rvval := rv.MapIndex(rvkey)
-		if rvval.IsValid() && rvval.Kind() == reflect.Interface {
+		if rvval.IsValid() && rvval.Kind() == reflect.Interface && ! rvval.IsNil() {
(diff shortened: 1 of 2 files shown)


pires:
@mechmind could you rebase your changes and @BurntSushi could you review for merge?

mechmind:
I rebased a branch but now i think about right way to implement such functionality. There are `UnmarshalTOML` interface that delegates unmarshalling to your handler and package `mapstructure` to fill structs from `map[string]interface{}`es. May be it is enough to replace this "schema-types" hack.

cespare:
I do not think we should add this feature. I haven't seen this behavior before in other reflect-based decoding libraries and I think it adds complexity to support a interface{}-heavy style of code that oughtn't be encouraged.

In particular, it seems to me that code using this feature would be better implemented using a struct rather than `map[string]interface{}`.

@BurntSushi please reopen and merge if you still think we should add this feature.

PR #67, opened 2015-01-19
Combine replaceUnicode and replaceEscapes to fix #66 and other problems ...

### LINKED ISSUE #66: Cannot round trip strings containing "\u"
Steps to reproduce:

1. Run this code:

```go
package main

import (
	"bytes"
	"fmt"

	"github.com/BurntSushi/toml"
)

func main() {
	data := map[string]string{
		"k": `c:\users`,
	}

	buf := &bytes.Buffer{}
	enc := toml.NewEncoder(buf)
	if err := enc.Encode(data); err != nil {
		fmt.Println(err)
		return
	}
	fmt.Println(buf.String())
	data2 := make(map[string]string)
	if _, err := toml.DecodeReader(buf, &data2); err != nil {
		fmt.Println(err)
		return
	}
	fmt.Println(data2)
}
```

I expect it to run without error and result in `data2` containing a copy of `data`.

Instead I get this output:

```
k = "c:\\users"

2015/01/16 11:05:00 BUG: Could not parse 'sers' as a hexadecimal number, but the lexer claims it's OK: strconv
.ParseUint: parsing "sers": invalid syntax
```

and the program aborts before completing.

ChrisHines:
...when parsing escape sequences.

The method of parsing unicode escapes in the existing code had several problems.

1. \\u was parsed the same as \u
2. \u005Cu was parsed the same as \u
3. \\u0075 was parsed the same as u
4. \\\u0075 was parsed the same as \u

Tests for the above sequences have been submitted in BurntSushi/toml-test#28

This PR adapts the code from encoding/json.unquoteBytes because the TOML spec says:

> the above string specification is the same as JSON's string definition, except that TOML requires UTF-8 encoding. This is on purpose

### DIFF
--- a/parse.go
+++ b/parse.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"time"
 	"unicode"
+	"unicode/utf16"
 	"unicode/utf8"
 )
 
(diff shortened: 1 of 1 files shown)


BurntSushi:
Thank you so much, but I ended up fixing this on my own. There's a lot of code in this PR, and it looks like it's a bit more complex than necessary for TOML because it tries to be fast. I ended up doing something a bit simpler but probably slower.

BurntSushi:
See https://github.com/BurntSushi/toml/commit/0eaa740fa3929209b5879375a0d6a9738f3bb826.

ChrisHines:
@BurntSushi Excellent! I expected you would want a simpler fix, but it was easier for me to crib the code from encoding/json in order to move past the issue and get back to my code. I agree that your fix is a better fit for TOML.

PR #79, opened 2015-04-09
fix parsing of UTF8+BOM files

tolbrino:


cespare:
I think that this library should not handle BOMs [for the same reason that encoding/json does not handle them](https://github.com/golang/go/issues/12254). It's easy enough to strip the BOM yourself, if your data is so afflicted.

PR #81, opened 2015-05-01
Support for omitempty, as well as tests for omitempty.

### LINKED ISSUE #73: Support "omitempty" like JSON and XML
It would be handy to tag my go struct with "omitempty" so that the toml writer does not write the value if it is empty. The go standard libraries `json` and `xml` have something similar.

Example:

```
type Settings struct {
    Port string `toml:"port,omitempty"`
}
```

bbuck:
This pull request will close #73 and add in support for omitempty.

If you want me to clean things up a bit, I can - this works fine but would not be the best structure to add in future "options" to field names.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -343,6 +343,21 @@ func (enc *Encoder) eStruct(key Key, rv reflect.Value) {
 			if keyName == "" {
 				keyName = sft.Name
 			}
+
+			opts := strings.Split(keyName, ",")
+			keyName = opts[0]
+			omitEmpty := false
+			if len(opts) > 1 {
+				for _, opt := range opts[1:] {
+					if opt == "omitempty" {
+						omitEmpty = true
+					}
+				}
+			}
+			if omitEmpty && isEmpty(sf) {
+				continue
+			}
+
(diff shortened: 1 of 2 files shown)


bbuck:
I've gone ahead and cleaned this up, turned into something a little more extendable.

I've also taken out the concept that zero numeric values are "empty" and thus would appear in `omitempty` fields when their values are `0`. To compensate, I added `omitzero` which would omit numeric fields when their values are `0`.

```go
type Sample struct {
	User  string `toml:"name"`
	Pass  string `toml:"password,omitempty"`
	Count int    `toml:"count,omitzero"`
}

s := Sample{"username", "", 50}
// Encode =>
// name = "username"
// count = 50
// 

s.Pass = "some password"
s.Count = 0
// Encode =>
// name = "username"
// password = "some password"
//
```

BurntSushi:
Looks great!

cespare:
It's unfortunate that omitzero is a separate concept from omitempty. This is an unnecessary break from encoding/json's behavior.

cespare:
See #120.

PR #82, opened 2015-05-19
Use correct name during decode with omit options

mjibson:


pires:
Need this. LGTM

cespare:
Thanks!

pires:
Finally! Thanks

PR #86, opened 2015-07-10
add toml tag "*" to allow append to current location

extrame:
Such as 

	type HekaPlugin struct {
		Type    string                 `toml:"type"`
		Args    map[string]interface{} `toml:"*"`
	}
	
Args with values :{"addr":"location","date":"today"}
Will Generate 

	[HekaPlugin]
	  type = "Type"
	  addr = "Location"
	  date = "Today" 
	  
instand of 
	
	[HekaPlugin]
	  type = "Type"
	  [HekaPlugin."*"]
	  	addr = "Location"
	  	date = "Today"

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -106,15 +106,15 @@ func (enc *Encoder) safeEncode(key Key, rv reflect.Value) (err error) {
 }
 
 func (enc *Encoder) encode(key Key, rv reflect.Value, indent ...bool) {
-	// Special case. Time needs to be in ISO8601 format.
-	// Special case. If we can marshal the type to text, then we used that.
-	// Basically, this prevents the encoder for handling these types as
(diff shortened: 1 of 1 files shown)


cespare:
I think we should not add this feature:

* You can already implement this yourself in a few different ways. For example, you could add all the fields of your struct into the args map and then encode that.
* It adds yet another struct tag option, giving more semantic meaning to stringly-typed code
* The symbol `*` isn't particularly self-explanatory, even compared with the existing option `-`

But most importantly, it's not clear how decoding into a struct using a `*` field would work. I think that this proposal would need to include a set of rules (and preferably code) for decoding so that data in this format can be round-tripped through the encoder and decoder.

cespare:
I'm closing this, for the reasons I gave above.

PR #87, opened 2015-07-23
Comment tag

dmitry-yakutkin:
Here's some functionality, which allows setting 'comment' tag for struct fields, which is recognized as toml-comment, and encoded properly.
Only some logic, which had made new lines in top level blocks was remade - it basically moved into eStruct() method.


cespare:
I don't think we should add more things into struct tags.

PR #90, opened 2015-08-03
Ignore field options (omitempty,omitzero) during Decode

### LINKED ISSUE #89: `omitempty` breaks Decode
With #81, toml supports the `omitempty` tag while encoding a struct. For example:

```go
type S struct {
    V string `toml:"value,omitempty"`
}
```

will skip writing "V" during `toml.Encode()`.

Unfortunately the reverse is not true. `omitempty` has no meaning for `toml.Decode()` and is treated as part of the field name. In the example above, `toml.Decode()` will attempt to decode the field named `value,omitempty` into "V" and will ignore the field named `value`.

chowey:
This fixes #89. Currently the only options are `omitempty` and `omitzero`, which have no significance while decoding, so these can simply be ignored.

I thought it best to use the same `getOptions` function so this stays unified in the future (if some additional options are added later).

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -349,6 +349,25 @@ func TestDecodeSizedInts(t *testing.T) {
 	}
 }
 
+func TestDecodeWithOmitEmpty(t *testing.T) {
+	type userpass struct {
+		User string `toml:"user"`
+		Pass string `toml:"password,omitempty"`
+	}
+	value := userpass{"Testing", "Filled"}
+	toml := `
+	user = "Testing"
+	password = "Filled"
+	`
+	var up userpass
+	if _, err := Decode(toml, &up); err != nil {
+		t.Fatal(err.Error())
+	}
+	if value != up {
+		t.Fatalf("Expected %#v but got %#v", value, up)
+	}
+}
+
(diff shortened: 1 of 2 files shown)


buaazp:
Yes, I meet this bug and @chowey 's PR fixed it.

If some fields has `,omitempty` option, they will be ignored when decoding.

For example, in the `_example/_example.go`, if I modified codes to:
```
type ownerInfo struct {
	Name string
	Org  string `toml:"organization,omitempty"`
	Bio  string
	DOB  time.Time
}
```
And then run it:
```
./_examples
Title: TOML Example
Owner: Tom Preston-Werner (, GitHub Cofounder & CEO
```
You can see the `Org` is blank now.

@BurntSushi plz merge these codes into master branch.

chowey:
Just noticed this is related to #82. I did the check in a different place.

titanous:
This could use a test to ensure that there isn't a regression.

chowey:
I added a test.

ayufan:
@BurntSushi Could you merge this?

cespare:
I went with #82 instead.

PR #94, opened 2015-09-17
Use utf8.ValidRune to check for valid Unicode

g2p:
Here, someone added an API that makes this easy.


cespare:
Heh, I independently did this in 001f7afe45039be55824867c4a245dbc90e8663b.

PR #95, opened 2015-09-30
bugfix: if a not pointer para is pass into decode(), an error will be…

ssor:
this confused me for a while

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -103,6 +103,10 @@ func (md *MetaData) PrimitiveDecode(primValue Primitive, v interface{}) error {
 // This decoder will not handle cyclic types. If a cyclic type is passed,
 // `Decode` will not terminate.
 func Decode(data string, v interface{}) (MetaData, error) {
+	inputStru := reflect.ValueOf(v)
+	if inputStru.Kind() != reflect.Ptr {
+		return MetaData{}, errors.New("The second para should passed in a pointer")
+	}
 	p, err := parse(data)
 	if err != nil {
 		return MetaData{}, err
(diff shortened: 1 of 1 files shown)


decode.go
```diff
@@ -103,6 +104,10 @@ func (md *MetaData) PrimitiveDecode(primValue Primitive, v interface{}) error {
 // This decoder will not handle cyclic types. If a cyclic type is passed,
 // `Decode` will not terminate.
 func Decode(data string, v interface{}) (MetaData, error) {
+	inputStru := reflect.ValueOf(v)
```
cespare: Name this `rv`

decode.go
```diff
@@ -103,6 +104,10 @@ func (md *MetaData) PrimitiveDecode(primValue Primitive, v interface{}) error {
...
+	if inputStru.Kind() != reflect.Ptr {
+		return MetaData{}, errors.New("The second para should passed in a pointer")
```
cespare: For the error, do

    e("Decode(non-nil %s)", reflect.TypeOf(v))

decode.go
```diff
@@ -103,6 +104,10 @@ func (md *MetaData) PrimitiveDecode(primValue Primitive, v interface{}) error {
...
+	}
```
cespare: While you're here, add a nil pointer check as well:

``` go
if rv.IsNil() {
        return MetaData, e("Decode(nil)")
}
```

cespare:
Timed out. I fixed the error myself in 375d86fa0c53e1d9b0fe823c31093f9cfe461b8d.

Issue #75, opened 2015-03-24
Support comments in encoding


synful:
Given that TOML supports comments, it would be nice if there were some way to omit TOML which included comments. My initial inclination was to implement the [TextMarshaler](http://godoc.org/github.com/BurntSushi/toml#TextMarshaler) interface, but it seems that doesn't work for key/value pairs (so you can't use it to omit a custom line of TOML).

My suggestion is that something be added which could be used to support this. I'm not sure what it would look like (a more flexible encoding mechanism? specific support for comments?), but it would be great to have.

synful:
Actually, here's an idea: struct field tags define comments. Something like:

```Go
type Person struct {
    Name string `toml:"name,the peron's name"`
    Age  string `toml:"age,the person's age"`
}
```

(the idea is that the comma separates the key name and the comment; the tag `toml:",comment"` would be valid)

tucnak:
Hugo +1 for this. @BurntSushi wanna push in?

gaillard:
+1

mffahey:
+1

blackheart01:
Would love to have this feature. +1 from me.

jbsmith7741:
This is a great idea. It would be nice to separate the comment into an additional tag. 
For example we could use something like desc (description) or com (comment)
``` golang 
type Person struct {
    Name string `toml:"name" desc:"the person's name"`
    Age  string `toml:"age" desc:"the person's age"`
}
```

PR #97, opened 2015-10-06
Support encoding comments in toml tag

echlebek:
Implements the feature suggested in #75. I'm not sure this is actually a good idea, but thought I'd share anyway.

A possible issue with this is that people who use comments might experience breakage if later on the library implements an option that matches their exact comment. That seems like a pretty remote possibility to me, since comments are usually partial sentences.

Another way to accomplish encoding comments would be the way encoding/xml does it, but I'm not particularly fond of that approach either tbh.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -424,6 +424,32 @@ name = "Rice"
 
 }
 
+func TestDecodeWithComment(t *testing.T) {
+	type elephant struct {
+		Age       int `toml:"age,Elephant age in years"`
+		TuskLen   int `toml:"tusk_length,Tusk length in cm"`
+		NWrinkles int `toml:"num_wrinkles,"`
+	}
+	e := elephant{}
+	var tomlBlob = `
+	# Hello world
+	age = 5
+
+	tusk_length = 10
+	num_wrinkles = 15
+	`
+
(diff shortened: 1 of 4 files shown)


DevelopSMM:
This branch works great for encoding.  However, decoding fails on all fields containing a comment.

echlebek:
@DevelopSMM thanks for the feedback. I've added a few commits that fix this problem.

I didn't bother rebasing or merging with tip, but would be happy to do so if there was interest in merging this branch.

cespare:
I lean towards not putting more things into the struct tags.

cespare:
I don't think we should go down the road of encoding arbitrary comments in struct tags.

SMEmp:
Absolutely makes sense, but do you have any solutions on how to handle comments then?  When working with config files it's nice to be able to document inline what each option is supposed to do.

cespare:
Comments are fully supported when decoding TOML. This issue is about encoding comments.

echlebek:
Yeah, this was never a good idea.

I think it's useful to be able to generate TOML with comments, (say for outputting default config files) but there's gotta be a better way. Maybe text/template.

Issue #100, opened 2015-11-16
Does Not Work Mashalling Struct Pointers


cretz:
Take a look at the following 4 tests:

```go
package toml_test

import (
	"testing"
	"bytes"
	"github.com/BurntSushi/toml"
	"encoding/json"
	"reflect"
)

type Foo struct {
	*Bar `json:"bar,omitempty" toml:"bar"`
	*Baz `json:"baz,omitempty" toml:"baz"`
}

type Bar struct {
	Text string `json:"text,omitempty" toml:"text"`
}

type Baz struct {
	Text string `json:"text,omitempty" toml:"text"`
}

func TestJsonUnmarshal(t *testing.T) {
	str := "{\"bar\":{\"text\":\"Some Text\"}}"
	foo := new(Foo)
	if err := json.Unmarshal([]byte(str), foo); err != nil {
		t.Fatal(err)
	}
	expected := &Foo{Bar: &Bar{Text: "Some Text"}}
	if !reflect.DeepEqual(expected, foo) {
		t.Fatal("Not equal")
	}
}

func TestJsonMarshal(t *testing.T) {
	foo := &Foo{Bar: &Bar{Text: "Some Text"}}
	byts, err := json.Marshal(foo)
	if err != nil {
		t.Fatal(err)
	}
	if string(byts) != "{\"bar\":{\"text\":\"Some Text\"}}" {
		t.Fatal("Not equal")
	}
}

func TestTomlUnmarshal(t *testing.T) {
	str := "[bar]\ntext = \"Some Text\""
	foo := new(Foo)
	if err := toml.Unmarshal([]byte(str), foo); err != nil {
		t.Fatal(err)
	}
	expected := &Foo{Bar: &Bar{Text: "Some Text"}}
	if !reflect.DeepEqual(expected, foo) {
		t.Fatal("Not equal")
	}
}

func TestTomlMarshal(t *testing.T) {
	foo := &Foo{Bar: &Bar{Text: "Some Text"}}
	buf := new(bytes.Buffer)
	if err := toml.NewEncoder(buf).Encode(foo); err != nil {
		t.Fatal(err)
	}
	if string(buf.Bytes()) != "[bar]\ntext = \"Some Text\"" {
		t.Fatal("Not equal")
	}
}
```

The last one breaks with:

        --- FAIL: TestTomlMarshal (0.00s)
        panic: reflect: call of reflect.Value.Type on zero Value [recovered]
[...]

PR #101, opened 2015-11-20
Don't continue trying to encode anonymous struct if it's a pointer ba…

ttacon:
…sed field and it's the Zero value


ttacon:
This addresses the panic in #100

cespare:
The issue was fixed in 2fe0945c12def7d89c84479fa3a21b287a70c227.

PR #103, opened 2015-11-23
Change true/false lexing to provide better error messages.

bjc:
This will consume an entire 'bareword' rather than trying to consume
either 'true' or 'false' in order to print more meaningful error
messages.

Previously, TOML such as:
  key = foo

Would print an error like:
  Expected "f", but found "o" instead.

Which is not very useful for people writing config files. Now it will
print an error like:
  Expected value but found 'foo' instead.

Which should be more useful.

### DIFF
--- a/lex.go
+++ b/lex.go
@@ -402,10 +402,6 @@ func lexValue(lx *lexer) stateFn {
 		}
 		lx.ignore() // ignore the "'"
 		return lexRawString
-	case r == 't':
-		return lexTrue
-	case r == 'f':
-		return lexFalse
 	case r == '-':
 		return lexNumberStart
 	case isDigit(r):
@@ -414,7 +410,11 @@ func lexValue(lx *lexer) stateFn {
 	case r == '.': // special error case, be kind to users
 		return lx.errorf("Floats must start with a digit, not '.'.")
 	}
-	return lx.errorf("Expected value but found %q instead.", r)
+
+	// Assume barewords are constants so we can emit better error
+	// messages.
+	lx.backup()
+	return lexBareword
 }
 
 // lexArrayValue consumes one value in an array. It assumes that '[' or ','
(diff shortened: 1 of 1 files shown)


lex.go
```diff
@@ -738,36 +738,29 @@ func lexFloat(lx *lexer) stateFn {
...
-	return nil
-}
 
-// lexTrue consumes the "rue" in "true". It assumes that 't' has already
-// been consumed.
-func lexTrue(lx *lexer) stateFn {
-	if fn := lexConst(lx, "true"); fn != nil {
-		return fn
+	s := string(rs)
+	if s == "true" {
```
cespare: switch s

lex.go
```diff
@@ -414,7 +410,11 @@ func lexValue(lx *lexer) stateFn {
 	case r == '.': // special error case, be kind to users
 		return lx.errorf("Floats must start with a digit, not '.'.")
 	}
-	return lx.errorf("Expected value but found %q instead.", r)
+
+	// Assume barewords are constants so we can emit better error
+	// messages.
+	lx.backup()
+	return lexBareword
```
cespare: I prefer how the code worked before, by only looking for a boolean when parsing a value beginning with `t` or `f`.

Instead of introducing a catch-all code path which ends up only handling booleans, it shouldn't be hard to instead change `lexConst` to give the improved error message by parsing a whole "constant" and comparing it to the target string.

lex.go
```diff
@@ -414,7 +410,11 @@ func lexValue(lx *lexer) stateFn {
 	case r == '.': // special error case, be kind to users
 		return lx.errorf("Floats must start with a digit, not '.'.")
 	}
-	return lx.errorf("Expected value but found %q instead.", r)
+
+	// Assume barewords are constants so we can emit better error
```
cespare: I'm not sure that "bareword" is better terminology than "constant".

cespare:
Hey @bjc, thanks for the change! I left a few comments for you. It's a nice improvement to the error.

cespare:
I went ahead and merged and did the cleanup myself.

PR #84, opened 2015-06-05
Make new destination slice when length doesn't match.

mrdg:
Fixes #83

PR #105, opened 2015-12-04
Overwrite existing slices when unmarshaling

### LINKED ISSUE #49: decoding table arrays need to be handled better
The following produces a runtime error:

```go
package main

import (
	"fmt"
	"log"

	"github.com/BurntSushi/toml"
)

type Server struct {
	Url  string
	Port int
}

type Host struct {
	Servers []Server
}

var blob = `
[Host]
[[Servers]]
Url = "http://burntsushi.net"
Port = 8080

[[Servers]]
Url = "http://github.com"
Port = 1234
`

func main() {
	config := Host{
		Servers: []Server{
			{
				Url:  "http://google.com",
				Port: 80,
			},
		},
	}
	if _, err := toml.Decode(blob, &config); err != nil {
		log.Fatal(err)
	}
	fmt.Printf("%#v\n", config)
}
```

This is because (I think) the default value. The decoder is probably checking if it's not nil, and then just assuming that it is the right size. It should be smarter. (Just appending isn't quite right either, because I think it should overwrite the default value. But I should probably check the behavior of the standard library decoders before implementing that.)

### LINKED ISSUE #99: For array field, parser performs a partial update instead of full replacement
Following code:

```
type Config struct {
	Hosts []string
}

func main() {
	cfg := Config{
		Hosts: []string{"aaa", "bbb"},
	}

	t := `Hosts = ["ccc"]`
	if _, err := toml.Decode(t, &cfg); err != nil {
		log.Fatal(err)
	}

	fmt.Printf("%v", cfg.Hosts)
}
```

Expect Result: `[ccc]`
Actual Result:  `[ccc, bbb]`

It seems parser performed a partial update instead of a full replacement. Is that possible to change parser's behaviour to full replacement?  Thanks!

SamWhited:
This brings the behavior inline with that of the standard library (eg.
encoding/json). This does change the behavior when the second slice is smaller (eg. before a new slice with 1 element would have overwritten just the first element of the existing slice, now you end up with a new slice). I think this is what most users will expect though (especially since it's what `encoding/json` and the like do).

Fixes #49
Fixes #99 

Eg. the following now works (which would have produced a panic before):

```go
package main

import (
	"fmt"
	"github.com/BurntSushi/toml"
)

func main() {
	blob := []byte(`
	test = ["aaa", "bbb"]

	[[table]]
	a = "test"
	`)

	type log struct {
		A string
		B string
	}

	type config struct {
		Table []log `toml:"table"`
		Test  []string
	}

	c := new(config)

	err := toml.Unmarshal(blob, &c)
	if err != nil {
		panic(err)
	}

	blob2 := []byte(`
	test = ["ccc"]

	[[table]]
	a = "test2"

	[[table]]
	a = "test3"
	`)

	err = toml.Unmarshal(blob2, &c)
	if err != nil {
		panic(err)
	}

	fmt.Printf("%+v\n", c)
}
```
Output:
```
&{Table:[{A:test2 B:} {A:test3 B:}] Test:[ccc]}
```

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -308,9 +308,7 @@ func (md *MetaData) unifySlice(data interface{}, rv reflect.Value) error {
 		return badtype("slice", data)
 	}
 	sliceLen := datav.Len()
-	if rv.IsNil() {
-		rv.Set(reflect.MakeSlice(rv.Type(), sliceLen, sliceLen))
-	}
+	rv.Set(reflect.MakeSlice(rv.Type(), sliceLen, sliceLen))
 	return md.unifySliceArray(datav, rv)
 }
 


cespare:
This was taken care of in #84.

PR #108, opened 2016-01-01
Fix endless loop in table name lexing

kezhuw:
lexTableNameStart returns itself without consuming any data in switch
case for whitespace.

Since TOML version v0.2.0 does not allow table name being surrounded by
whitespaces, simply dropping whitespace case solves this.

Following are examples that triggered endless loops. This commit
complains on these.

```toml
[   table1]

[table2.  ]
```


BurntSushi:
Great, thank you!

PR #112, opened 2016-02-19
Bugfix: update check for inaccessible fields for Go 1.6

stapelberg:
As per https://golang.org/doc/go1.6#reflect


BurntSushi:
Great, thanks!

PR #119, opened 2016-03-03
add bool empty option

jackspirou:
Boolean values should be recognized as empty if false. Do it like the standard library does: https://github.com/golang/go/blob/master/src/encoding/json/encode.go#L286-L287

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -494,6 +494,8 @@ func isEmpty(rv reflect.Value) bool {
 		if rv.Len() == 0 {
 			return true
 		}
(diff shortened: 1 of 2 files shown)


cespare:
Looks good. Please add a test.

jackspirou:
@cespare tests add for boolean values use in the `isEmpty` function.  git commits squashed.

cespare:
Thanks!

jackspirou:
@cespare np.  Thanks for the quick merge :)

PR #132, opened 2016-04-26
Support all TOML datetimes

cespare:
Depends on #131.


cespare:
This isn't going to pass toml-test because of https://github.com/BurntSushi/toml-test/issues/35.

PR #134, opened 2016-05-05
Support decoding a singular element into a slice array

jsternberg:
This teaches the TOML decoder how to decode a singular element into a
slice array. If you have a type like this:

    type Config struct {
        Inputs []InputConfig `toml:"inputs"`
    }

    type InputConfig struct {
        Name string `toml:"name"`
    }

You can now specify the inputs as either a single element or an array
inside of the TOML:

    [inputs]
      name = "foobar"

This will create a singular slice with one element and decode the
mapping into the struct.

This is useful when you previously only allowed a single element and you
want to expand the config to allow multiple inputs to be selected.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -323,6 +323,13 @@ func (md *MetaData) unifySlice(data interface{}, rv reflect.Value) error {
 	if datav.Kind() != reflect.Slice {
 		if !datav.IsValid() {
 			return nil
+		} else if _, ok := data.(map[string]interface{}); ok {
+			rv2 := reflect.MakeSlice(rv.Type(), 1, 1)
+			if err := md.unifyStruct(data, rv2.Index(0)); err != nil {
+				return err
+			}
+			rv.Set(rv2)
+			return nil
 		}
 		return badtype("slice", data)
 	}


cespare:
I don't think we should add this special case. It's better to be strict about types.

I can think of a couple of ways you can support multiple versions of your config. Here's one that uses toml.Primitive:

``` go
package main

import (
	"fmt"
	"log"

	"github.com/BurntSushi/toml"
)

type Config struct {
	Inputs []InputConfig
}

type InputConfig struct {
	Name string
}

func decode(s string) (*Config, error) {
	var c struct {
		Inputs toml.Primitive
	}
	meta, err := toml.Decode(s, &c)
	if err != nil {
		return nil, err
	}
	var inputs []InputConfig
	switch meta.Type("inputs") {
	case "Hash":
		var input InputConfig
		if err := meta.PrimitiveDecode(c.Inputs, &input); err != nil {
			return nil, err
		}
		inputs = append(inputs, input)
	case "ArrayHash":
		if err := meta.PrimitiveDecode(c.Inputs, &inputs); err != nil {
			return nil, err
		}
	}
	return &Config{Inputs: inputs}, nil
}

func main() {
	const in0 = `
[inputs]
name = "one"
`

	const in1 = `
[[inputs]]
name = "two"
`

	for _, s := range []string{in0, in1} {
		c, err := decode(s)
		if err != nil {
			log.Fatal(err)
		}
		fmt.Printf("%#v\n", c)
	}
}
```

output:

```
&main.Config{Inputs:[]main.InputConfig{main.InputConfig{Name:"one"}}}
&main.Config{Inputs:[]main.InputConfig{main.InputConfig{Name:"two"}}}
```

Issue #76, opened 2015-03-25
Proposal: json-style custom marshalling


joshlf:
In the [json](http://golang.org/pkg/encoding/json/) package, encoding a type which implements the [Marshaler](http://golang.org/pkg/encoding/json/#Marshaler) interface will cause the encoder to call the MarshalJSON method in order to encode the value. Notably, this succeeds so long as the produced output is valid JSON.

This package allows similar functionality, but restricts it to string values. It would be very nice if this were expanded so that any valid TOML were supported. To provide some motivation: @ezrosent and I ran into this issue today in trying to encode numerical values, and ended up having to use a [horrible hack](https://github.com/synful/kudos/commit/63de72d2490c22a99d33711b2c7a7d4a0bd88724) to work around this limitation :P

cespare:
I think we should do this.

Issue #63, opened 2014-12-17
toml.Primitive should encode correctly


chowey:
If I have a struct like this:

``` go
type SomeStruct struct {
    PreserveData toml.Primitive
    SomeData *MyData
}
```

I would expect that I can read a toml file, decode it to a `SomeStruct`, change the `SomeData`, and encode it back into a file with the `PreserveData` basically staying the same. Instead, it encodes to nothing.

Can we have a `toml.Primitive` encode back into whatever it used to be? Sort of like how `json.RawMessage` works?

cespare:
If we do json-style custom encoders (`toml.Marshaler`, basically -- see #76) then this can cleanly be implemented by adding the appropriate methods to `toml.Primitive`.

Otherwise we have to special-case it in the encoder (as it's already special-cased in the decoder).

arp242:
> If we do json-style custom encoders (toml.Marshaler, basically -- see #76) then this can cleanly be implemented by adding the appropriate methods to toml.Primitive.
>
> Otherwise we have to special-case it in the encoder (as it's already special-cased in the decoder)

Considering this hasn't been fixed almost 7 years after this PR was first opened and that the special case is just three lines of code, fixing it like this now sound reasonable. It solves actual issues people are facing and can always improve later on when that gets implemented.

I merged this via #281, which adds a comment for the above so it's less likely to get forgotten and improves the test that was added for this.

PR #137, opened 2016-05-23
add Marshaler interface

kovetskiy:
Hi @BurntSushi, I've noticed that there is no nice way to handle Marshal/Encode operation unless implementing `encoding.TextMarshaler` interface, which can't be used if struct should be also marshalable to JSON  (json libraries also looks for `encoding.TextMarshaler`),

so I've added `Marshaler` interface with following signature:

```
type Marshaler interface {
    MarshalTOML() ([]byte, error)
}
```

Also I've added various tests for this feautre.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -38,6 +38,12 @@ var quotedReplacer = strings.NewReplacer(
 	"\\", "\\\\",
 )
 
+// Marshaler is the interface implemented by objects that can marshal
+// themselves to a TOML representation.
+type Marshaler interface {
+	MarshalTOML() ([]byte, error)
+}
+
 // Encoder controls the encoding of Go values to a TOML document to some
 // io.Writer.
 //
@@ -106,6 +112,10 @@ func (enc *Encoder) safeEncode(key Key, rv reflect.Value) (err error) {
 }
 
 func (enc *Encoder) encode(key Key, rv reflect.Value) {
+	if _, ok := rv.Interface().(Marshaler); ok {
+		rv = marshalDecode(rv)
+	}
+
 	// Special case. Time needs to be in ISO8601 format.
 	// Special case. If we can marshal the type to text, then we used that.
 	// Basically, this prevents the encoder for handling these types as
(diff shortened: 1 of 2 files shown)


kovetskiy:
@BurntSushi ping

encode.go
```diff
@@ -114,6 +120,9 @@ func (enc *Encoder) encode(key Key, rv reflect.Value) {
 	case time.Time, TextMarshaler:
 		enc.keyEqElement(key, rv)
 		return
+
+	case Marshaler:
```
cespare: If a user-defined type is a Marshaler and a TextMarshaler, the former should take precedence.

kovetskiy:
@cespare I've force pushed changes with Marshaler precedence.

kovetskiy:
@BurntSushi @cespare ping

kovetskiy:
@BurntSushi @cespare any reaction?

cespare:
Fixes #76 
Related to #63

cespare:
@kovetskiy sorry, been busy. I'll try to take a look this weekend.

kovetskiy:
@cespare @BurntSushi ping

kovetskiy:
@cespare @BurntSushi ping

tucnak:
jeez @kovetskiy it must hurt lol

kovetskiy:
@tucnak yeah, such maintainers

encode.go
```diff
@@ -537,6 +551,21 @@ func encPanic(err error) {
 	panic(tomlEncodeError{err})
 }
 
+func marshalDecode(rv reflect.Value) reflect.Value {
```
cespare: You're already calling `rv.Interface().(Marshaler)` in the caller; pass in a `Marshaler` rather than a `reflect.Value` and avoid the duplicate work.

cespare:
I was hoping that adding Marshaler could be a solution for #64, #75, and other cases where there's no way to output a particular encoded value.

For example, I was thinking that if you have a MarshalTOML method like

``` go
func (T) MarshalTOML() ([]byte, error) {
	return []byte(`# hello
a = """
1
2
3"""
b = '\n'`), nil
}
```

then that would yield a TOML encoding with a comment, a multi-line string, and a literal string:

``` toml
# hello
a = """
1
2
3"""
b = '\n'
```

but since your approach is to decode and then encode the value, we instead get

``` toml
a = "1\n2\n3"
b = "\\n"
```

I agree that we need to decode the value to check that it's valid TOML, but I think we should emit the original encoding. WDYT?

siddharthist:
@kovetskiy Any plans to address the comments here?

kovetskiy:
@siddharthist Yes, I've missing comment. Will fix soon.

PR #139, opened 2016-06-16
fix 'panic: reflect.Value.Interface: cannot return value obtained fro…

mohae:
…m unexported field or method'

This fixes a panic that would occur if a struct field used a pointer by checking to see if the element is both addressable and settable by adding a `CanSet()` check.

Prior to this, when unmarshaling a TOML config in one of my apps, I would get:

    panic: reflect.Value.Interface: cannot return value obtained from unexported field or method\

on `github.com/BurntSushi/toml/decode.go:483`

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -531,7 +531,7 @@ func TestDecodeFloats(t *testing.T) {
 			continue
 		}
 		if x.N != tt.want {
-			t.Errorf("Decode(%q): got %d; want %d", input, x.N, tt.want)
+			t.Errorf("Decode(%q): got %f; want %f", input, x.N, tt.want)
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


cespare:
Can you add a test case?

mohae:
oh, I thought I had added it to decode_test.go; I should have checked the diff the formatting change is there (which i didn't mean to include, but it does correct a lint message.)

Sorry about forgetting to make sure the test case was there; I'll add it shortly.

decode_test.go
```diff
@@ -531,7 +531,7 @@ func TestDecodeFloats(t *testing.T) {
 			continue
 		}
 		if x.N != tt.want {
-			t.Errorf("Decode(%q): got %d; want %d", input, x.N, tt.want)
+			t.Errorf("Decode(%q): got %f; want %d", input, x.N, tt.want)
```
cespare: Shouldn't these both be `%f`?

mohae: yes, that was updated too.  Thanks for pointing it out.

mohae:
ok, I finally straightened things out.  Hopefully you can squash and merge to eliminate the noise I introduced, otherwise I can submit a cleaner request.

Thanks!

cespare:
Thanks for this fix. I opted for a slightly different change and a simpler test in d8fdfc4d09e2d991fc405efa6c8dcd4afcf02cda.

PR #143, opened 2016-07-12
travis: update to current go versions

martinlindhe:


cespare:
Hi @martinlindhe, the build is broken because of BurntSushi/toml-test#35, so this won't fix it.

Additionally, this library includes some specific code to work with old versions of Go, so I'd prefer not to remove those from the CI tests.

martinlindhe:
Thanks for pointing me in the right direction. This patch was indeed in attempt to resolve the failing test as seen in travis.

PR #144, opened 2016-07-17
Run .travis.yml against latest Go versions

kevinburke:
Previously the test runner only ran against Go 1.2 and tip (Go 1.7rc1), there
are 4 versions that have been released between those.

Another approach would be to only test against the latest two releases, Go 1.5
and Go 1.6 - let me know.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ language: go
 go:
   - 1.1
   - 1.2
+  - 1.3
+  - 1.4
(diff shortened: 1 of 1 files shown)


kevinburke:
Note the tests currently fail, they should hopefully pass after https://github.com/BurntSushi/toml-test/pull/41

BurntSushi:
Yup, looks good, thank you!

PR #166, opened 2017-03-21
break lexBool loop on ',' to allow decoding []bool

maclav3:
lexBool would break on decoding `key = [true, false]`, because the loop would go up to `true,` and then miss both cases.

### DIFF
--- a/lex.go
+++ b/lex.go
@@ -840,7 +840,7 @@ func lexBool(lx *lexer) stateFn {
 	var rs []rune
 	for {
 		r := lx.next()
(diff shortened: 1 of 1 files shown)


cespare:
@maclav3 thanks very much for bringing this bug to our attention.

There were some other cases besides comma: for instance, in your example, the `]` that comes after `false` would become part of that token. The inline table terminator, `}`, is another exception.

I ended up fixing this using a different workaround in b26d9c308763d68093482582cea63d69be07a0f0.

PR #172, opened 2017-04-26
Fix override map element value during decode

bgaifullin:
I have map, that has structures as elements.
```go
type LogConfig struct {
     Filename string  `toml:"file"`
     Level       string  `toml:"level"`
}

type Config struct {
    Loggers map[string]*LogConfig `toml:"log"`
}

fun NewConfig() *Config {
    return &Config{
         Loggers: map[string]*LogConfig {
                 "error": &LogConfig{Level: "error"},
                 "access": &LogConfig{Level: "info"},
        }
}

cfg := NewConfig()
toml.DecodeFile("config.toml", &cfg)
```

And If I do not specify the both "level and file" in config, I lose the defaults with I specify on creating Config.
 
I expected that level will keep origin value if it is not specified in config.toml

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -290,16 +290,20 @@ func (md *MetaData) unifyMap(mapping interface{}, rv reflect.Value) error {
 	for k, v := range tmap {
 		md.decoded[md.context.add(k).String()] = true
 		md.context = append(md.context, k)
-
-		rvkey := indirect(reflect.New(rv.Type().Key()))
-		rvval := reflect.Indirect(reflect.New(rv.Type().Elem()))
+		rvkey := reflect.ValueOf(k)
+		rvval := reflect.Indirect(rv.MapIndex(rvkey))
+		created := false
+		if !isComposite(rvval) {
+			created = true
+			rvval = reflect.Indirect(reflect.New(rv.Type().Elem()))
+		}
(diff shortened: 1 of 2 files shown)


bgaifullin:
@BurntSushi, please review PR

BurntSushi:
What is the behavior of other reflection based packages such as `encoding/json`?

bgaifullin:
The encoding/json keeps values, which is defined in original map, but does not defined in json.
there is example: https://play.golang.org/p/UJoOfzx8-w

bgaifullin:
@BurntSushi, please check example with encoding/json.

bgaifullin:
ping

bgaifullin:
@BurntSushi Could you please review this PR

bgaifullin:
up

cespare:
Hi @bgaifullin,

I think that we behave the same way that encoding/json does here. Your JSON example is showing how an existing map is preserved, but your TOML example seems to show that the LogConfig struct field defaults aren't preserved.

Here's a runnable comparison:

https://play.golang.org/p/d6t50gCppF

If you disagree, please give a complete runnable code example that shows the difference.

bgaifullin:
Hi @cespare
You are right. I made wrong comparison.  
I will close the PR

PR #179, opened 2017-06-05
Support maps with interface{} type keys

### LINKED ISSUE #178: Can't encode a map with interface{} keys, even if the keys are strings
Sometimes I hit situations where I have a map like:

```go
m := map[interface{}]interface{}{"foo": "bar"}
```

IMO, this _should_ encode cleanly to TOML:

```toml
foo = "bar"
```

But when I try this, I get an error: `toml: cannot encode a map with non-string key type`.

I encounter maps like this most often when I decode YAML with libraries like [`gopkg.in/yaml.v2`](https://github.com/go-yaml/yaml/tree/v2), and unfortunately since YAML supports more than just `string` as a key type, they have to use `interface{}` as the key type.

I think this is fixable by simply attempting to coerce the key (here: https://github.com/BurntSushi/toml/blob/master/encode.go#L276). i.e. `mapKey.Interface().(string)` instead of `mapKey.String()`.

I'll submit a PR to do this if my approach makes sense.

hairyhenderson:
Fixes #178 

This allows `map[interface{}]interface{}` to be encoded, _if_ each and every key can be coerced to be a `string`. See #178 for more explanation.

Signed-off-by: Dave Henderson <dhenderson@gmail.com>

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -265,7 +265,7 @@ func (enc *Encoder) eMapOrStruct(key Key, rv reflect.Value) {
 
 func (enc *Encoder) eMap(key Key, rv reflect.Value) {
 	rt := rv.Type()
-	if rt.Key().Kind() != reflect.String {
+	if rt.Key().Kind() != reflect.String && rt.Key().Kind() != reflect.Interface {
 		encPanic(errNonString)
 	}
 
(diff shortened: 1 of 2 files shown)


hairyhenderson:
cc @BurntSushi - is it possible to merge this? Or is there anything else you need me to do to get this merged?

arp242:
Closing this per the comment on https://github.com/BurntSushi/toml/issues/178#issuecomment-856499803

PR #182, opened 2017-06-22
Allow '+' in Datetime lexing

anthonyfok:
See gohugoio/hugo/issues/3626, where a Hugo user in China, with a positive UTC offset of `+0800`, encountered the following error when using an unquoted `date = 2017-06-22T13:39:12+08:00`:

> ERROR 2017/06/22 13:41:27 failed to parse page metadata for "post/former.md": Near line 2 (last key parsed ''): expected a top-level item to end with a newline, comment, or EOF, but got '+' instead

This appears to be an easy fix: Adding `+` as an allowed character in lexDatetime() seems to do the trick.  Please review.  Many thanks!

(See also gohugoio/hugo/issues/3628)


bep:
@nathany @cespare this issue creates some noise from the Hugo users in China etc., I plan to do a patch release with some fixes on Monday, and it would be great if this could be merged before then. It looks uncontroversial to me. Thanks!

BurntSushi:
The TOML spec says that the datetime format should use RFC 3339, and [RFC 3339 says that `+` is part of the syntax](https://tools.ietf.org/html/rfc3339#section-5.6), so this makes sense to me! Thanks!

@bep Out of curiosity, why is the switch to this lib temporary?

bep:
>@bep Out of curiosity, why is the switch to this lib temporary?

We used this lib before, but switched for "political reasons" some time ago. I think Kubernetes used one of the @spf13 libs (Viper?), but then some Google lawyer disliked this repo's license, so they asked us to change to a TOML lib with a "better" license. Not my finest decision, as it has given me more work/issues than I like to. The reason we switched back "temporary" now was that this lib proved to be much faster than the other one. And the temporary thing is mainly motivated by not having to TOML libs as dependencies.

BurntSushi:
@bep Gotya. Yeah, multiple Google employees have contacted me over the issue. I've opened a new issue to relicense: #183

decode_test.go
```diff
@@ -16,6 +16,8 @@ age = 250
 andrew = "gallant"
 kait = "brady"
 now = 1987-07-05T05:45:00Z
+nowCanada = 2017-06-22T02:14:36-06:00
+nowChina = 2017-06-22T16:15:21+08:00
```
nathany: It looks like the simple struct below should include a few more fields and answer should do a comparison of these to complete the testing.

nathany: Hah. Sorry, I didn't see that this was already merged before doing a review. 🤦‍♂️

anthonyfok: > It looks like the simple struct below should include a few more fields and answer should do a comparison of these to complete the testing.

Hi Nathan!  Yes, you are absolutely right!  I was lazy and just wanted to get the issue solved in Hugo and ended up not really examining decode_test.go at all.  (really quick and dirty)

Thanks for catching that!  I will make another pull request which will hopefully complete the test.

BurntSushi: @anthonyfok Thanks!

PR #184, opened 2017-06-23
relicense to MIT

### LINKED ISSUE #183: switch to MIT license
The WTFPL causes people to not want to use this library. Historically, I've been against changing the license mostly because it's just a pain to do so. Ideally, we'd like to get a sign off from all people that have made non-trivial contributions. (And if we can't, we either need to stop the re-licensing process or remove/rewrite those contributions. I'm not particularly inclined to the do latter, but if we can do the former, than we should.)

So, this issue is going to track the relicensing progress. I'll start by listing some of the major contributors, where I interpreted "major" as "100 lines of code or more."

* [x] @BurntSushi
* [x] @cespare 
* [x] @sqs
* [x] @rjeczalik
* [x] @ttacon
* [x] @bbuck
* [x] @halostatue
* [x] @sethwklein

BurntSushi:
Closes #183

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,14 +1,21 @@
-            DO WHAT THE FUCK YOU WANT TO PUBLIC LICENSE
-                    Version 2, December 2004
+The MIT License (MIT)
 
- Copyright (C) 2004 Sam Hocevar <sam@hocevar.net>
+Copyright (c) 2015 Andrew Gallant
 
- Everyone is permitted to copy and distribute verbatim or modified
- copies of this license document, and changing it is allowed as long
- as the name is changed.
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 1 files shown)


ryanwalls:
@BurntSushi Just curious.  Why COPYING instead of LICENSE?  Our automated tools for license checking don't pick up the COPYING file... so I have to manually come here.  Then I couldn't find the license file, so I search issues to find the issue about switching to MIT #183.  Then I had to look at this PR that closed that to then find that the file was called COPYING.  Whew.  :)

BurntSushi:
@ryanwalls `COPYING` is a standard name for the license file.

ryanwalls:
@BurntSushi Ah.  Good to know.  Didn't realize it was standard.  Will send a PR to the licensing tool I'm using.  https://github.com/pivotal/LicenseFinder

ryanwalls:
@BurntSushi Since this has come up before, see https://github.com/kubernetes/kubernetes/pull/31024, it would seem a small change to standardize.  But I understand if it is a low priority.

PR #185, opened 2017-06-26
Complete the testing for decoding datetime with UTC offset

anthonyfok:
Kudos to @nathany who wrote:

> It looks like the simple struct below should include a few more fields
> and answer should do a comparison of these to complete the testing.

in his code review for Pull Request #182.


BurntSushi:
@anthonyfok Thanks so much!

PR #189, opened 2017-08-18
Decode primitive back to original #63

outcoldman:
Should fix https://github.com/BurntSushi/toml/issues/63
Special case for toml.Primitive when Encode


arp242:
Merged this PR via #189, which rebases on master to solve the conflicts and improves the test case a bit. Thanks!

PR #193, opened 2017-09-28
Add ability to encode comments with desc tag

jbsmith7741:
Resolve issue #75 by adding a new tag 'desc' to describe variables.

example struct 
``` golang
type Person struct {
	Name string `desc:"My Name"`
	Age  int    `desc:"My Age"`
	Dog  Dog    `desc:"My dog"`
}

type Dog struct {
	Name  string `desc:"My Dog"`
	Breed string `desc:"Dog's breed"`
	Pet  Bug    `desc:"Dog's pet"`
}

type Bug struct {
	Weight float64 `desc:"in ounces"`
}
```
encodes to 
``` toml
# My Name
Name = ""
# My Age
Age = 0
# My dog
[Dog]
  # My Dog
  Name = ""
  # Dog's breed
  Breed = ""
  # Dog's pet
  [Dog.Pet]
    # in ounces
    Weight = 0.0
```

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -377,7 +377,7 @@ func (enc *Encoder) eStruct(key Key, rv reflect.Value) {
 }
 
 func (enc *Encoder) addComment(key Key, tag reflect.StructTag) {
-	if s, ok := tag.Lookup("desc"); ok {
+	if s := tag.Get("desc"); s != "" {
 		enc.wf("%s#%s\n", enc.indentStr(key), s)
 		enc.hasWritten = false
 	}
--- a/encode.go
+++ b/encode.go
@@ -368,11 +368,19 @@ func (enc *Encoder) eStruct(key Key, rv reflect.Value) {
 				continue
 			}
 
+			enc.addComment(key.add(keyName), sft.Tag)
 			enc.encode(key.add(keyName), sf)
 		}
 	}
 	writeFields(fieldsDirect)
 	writeFields(fieldsSub)
+}
+
+func (enc *Encoder) addComment(key Key, tag reflect.StructTag) {
+	if s, ok := tag.Lookup("desc"); ok {
+		enc.wf("%s#%s\n", enc.indentStr(key), s)
+		enc.hasWritten = false
+	}
 }
 
 // tomlTypeName returns the TOML type name of the Go value's type. It is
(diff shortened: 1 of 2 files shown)


encode.go
```diff
@@ -368,13 +368,21 @@ func (enc *Encoder) eStruct(key Key, rv reflect.Value) {
...
 			enc.encode(key.add(keyName), sf)
 		}
 	}
 	writeFields(fieldsDirect)
 	writeFields(fieldsSub)
 }
 
+func (enc *Encoder) addComment(key Key, tag reflect.StructTag) {
+	if s := tag.Get("desc"); s != "" {
+		enc.wf("%s#%s\n", enc.indentStr(key), s)
```
zachbadgett: Maybe we can add a space after the #

silasdavis:
Would be useful to have this merged. If you want to have a single struct to define your application's config then it is nice to be able to maintain a two-way mapping between your canonical 'init' config with helpful comments and from modified config back to your struct. Without this you end up with the overhead of a manually assembled example/default config with comments and run the risk of them getting out of sync.

gaillard:
Would also be useful to decode the comments into something accessible, perhaps that means not using field tags?

jbsmith7741:
FYI: I've added this feature in another fork if anyone is interested https://github.com/hydronica/toml

arp242:
I'm not a huge fan of using struct tags for this. Struct tags are limited in what you can put in there, it's a static string you can never modify (i.e. what if I want to put dynamic data in there?) and it just looks ugly for long strings IMO.

Instead, we could do something like:

```
type Commenter interface {
    Comment() string
}
```

And then you can use something like:

```
type X struct {
    WithComment strCmt
}

type strCmt string

func (strCmt) Comment() string { return "Yo, a comment!" }
```

Or ... something like that. I'm not sure. It's a bit more boilerplate, but it's a lot more flexible.

How do other encoders solve this problem, if at all? I can't really find an example of another encoder which does something like this.

arp242:
I'll close this for now as I almost certainly don't want to use struct tags for this. I have some ideas for a better solution, as mentioned, but I need to prototype a few things to see what does and doesn't work. And before I take this on I want to fix some other things, like supporting all of TOML 1.0, fixing a few remaining bugs, and improving the errors a wee bit.

I absolutely want to add this feature on the short term (i.e. weeks, not years), just need to think of the best way to do it. You can track #75 for updates.

PR #198, opened 2017-10-18
Unify indirect value of interface type instead of unifyAnything() in unify()

ktr0731:
Hi!  
I think `unify` method should call `unify` with indirect(rv) when `rv`'s type is interface.


BurntSushi:
Is this causing a bug somewhere?

ktr0731:
Oops... 
I will review my changes again.

ktr0731:
I close this PR temporarily.

PR #199, opened 2017-11-02
update: more accurate error message

jayce:
Now, you will get message:
```
$ go test
--- FAIL: TestDecodeSimple (0.04s)
        decode_test.go:53: toml: parsed 'My.Cats.plato': cannot load TOML value of type int64 into a Go string
FAIL
exit status 1
FAIL    _/D_/Github/toml        1.067s
```

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -11,7 +11,7 @@ import (
 )
 
 func e(format string, args ...interface{}) error {
-	return fmt.Errorf("toml: "+format, args...)
+	return fmt.Errorf(format, args...)
 }
 
 // Unmarshaler is the interface implemented by objects that can unmarshal a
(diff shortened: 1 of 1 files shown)


decode.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 func e(format string, args ...interface{}) error {
-	return fmt.Errorf("toml: "+format, args...)
+	return fmt.Errorf(format, args...)
```
BurntSushi: Does this result in removing the `toml` prefix from other error messages?

jayce: Yes, I will fix this question.

arp242:
It's a little bit unclear to me what exactly this changes; what was the output before, and what is it after this patch? Could you clarify this? A test for the new behaviour would be nice as well.

(old PR, so it's okay if you lost interest 😅)

arp242:
Closing for now due to lack of feedback. Feel free to get back later and we can reopen. Thanks!

PR #202, opened 2017-11-26
New error message for unescaped backslashes

axxelG:
My proposal for a better error message


arp242:
To be honest I think the current message `expected four hexadecimal digits after '\u' but got %q instead` is clear enough; maybe a little bit more usage guidance about escaping `\` to `\\` can be added, but I'd expect most people already know about this but just forgot in a copy/paste or the like?

Also, the way it's formatted now means there is no space between `instead` and `Maybe`, and the way it's phrased suggests what `/` is an alternative for `\`, so it certainly can't be merged as-is based on that.

PR #205, opened 2017-12-14
Unpack primitive data during encoding

### LINKED ISSUE #63: toml.Primitive should encode correctly

jkakar:
The encoder was treating `Primitive` values as plain old structs, which was resulting in incorrect encoding. The fix here special cases handling of primitive values to extract the internal data so that it can be rendered correctly.

@praboud and I debugged this issue together.

Fixes #63.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -1,6 +1,7 @@
 package toml
 
 import (
+	"bytes"
 	"fmt"
 	"log"
 	"math"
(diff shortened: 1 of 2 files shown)


jkakar:
Hmm, the fix in 083038e is cleaner. Will close this out for now and wait for that to land.

PR #216, opened 2018-03-12
Support meaningful error message for non-string map key decoding

motomux:
Currently, the package panics with a raw error message while decoding non-string map key.

```
panic: reflect: call of reflect.Value.SetString on int Value
```

This fix lets the package panic with a meaningful error message.

```
panic: toml: cannot decode map with non-string key
```

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -1,6 +1,7 @@
 package toml
 
 import (
+	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


arp242:
Thanks, I fixed this in a better way in #280; this really shouldn't panic at all but just return an error.

PR #218, opened 2018-03-17
Support fractional seconds in datetimes

sgarciac:


sgarciac:
note: the lexer already accepts fractional seconds

PR #230, opened 2018-08-14
switch the licenses of cmd/ to be the same as the root

gregwebs:
It seems like this was just missed when addressing #183. It showed up in my license scanner (fossa.io).


BurntSushi:
Indeed. Thanks!

Issue #217, opened 2018-03-14
Add line/column reporting to errors


CapacitorSet:
In brief, it would be nice to have the decoder report the line and column where an error occurred. As it is now, it reports errors like `toml: cannot load TOML value of type map[string]interface {} into a Go slice`, which while meaningful are not too helpful for the user.

grzkv:
This would be great to have.

PR #245, opened 2019-03-27
Update decode error to print data value

saschagrunert:
Hey, this small change would increase the debugability on errors during decoding. Now we would have a small hint which value causes the actual issue. WDYT?

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -505,5 +505,6 @@ func isUnifiable(rv reflect.Value) bool {
 }
 
 func badtype(expected string, data interface{}) error {
-	return e("cannot load TOML value of type %T into a Go %s", data, expected)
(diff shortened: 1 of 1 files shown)


arp242:
I think the idea is good, but the problem is that the value that it prints may not necessarily be the value in the TOML file. I might have `eat = 0xbeef`, but this will then print as `48879` with this, which isn't what's in the TOML file at all and is probably more confusing than helpful. There are a few other cases where this can happen (e.g. dates). Also, sometimes the value may be very long strings (although trimming that is easy).

I think the way forward is to 1) make sure that the data *as it appears in the file* is added, and/or 2) add line/column data to the error (there's an issue for that already: #217).

Either way, I think getting this right is quite a bit more complex than what's proposed in this patch right now and has the potential to be *even more* confusing, so I'll close this for now.

PR #247, opened 2019-07-20
Replaced range loop with a copy

muesli:
Could potentially also just copy here.

### DIFF
--- a/parse.go
+++ b/parse.go
@@ -458,10 +458,8 @@ func (p *parser) setValue(key string, value interface{}) {
 // Note that if `key` is empty, then the type given will be applied to the
 // current context (which is either a table or an array of tables).
 func (p *parser) setType(key string, typ tomlType) {
-	keyContext := make(Key, 0, len(p.context)+1)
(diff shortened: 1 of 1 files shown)


muesli:
Update: followed my own suggestion and pushed a newer version which uses copy instead of append.

arp242:
The behaviour on this isn't exactly identical: in the original version it would allocate a `Key` (`[]string`) slice with a len of 0 and cap of `len(p.context)+1`, the end result after the loop was a slice with the len one less than the cap.

In your version the len and cap are the same, and the result will `toml.Key{"asd", ""}` instead of `toml.Key{"asd"}`.

The reason it's `+1` is it can optionally append one more entry below this: `if len(key) > 0 { // allow type setting for hashes`.

The tests still pass so it doesn't seem to matter hugely, but unless there's some concrete benefit to using `copy()` (faster? maybe?) I think the existing version is better, so I'll close this for the time being, but we can reopen and fix it if it does turn out it improves things in some way.

PR #251, opened 2019-09-20
track index relative to recursion depth in `addFields` calls

### LINKED ISSUE #250: Problem encoding nested anonymous structs
I'm having a problem round-tripping a structure containing nested anonymous structs.

https://play.golang.org/p/uMig0i_5GmK

```
package main

import (
	"bytes"
	"fmt"
	"log"

	"github.com/BurntSushi/toml"
)

type (
	A struct {
		A string
	}

	B struct {
		B string
	}

	C struct {
		C string
	}

	BC struct {
		B
		C
	}

	Foo struct {
		A
		BC
	}
)

func main() {
	foo := &Foo{
		A: A{
			A: "a",
		},
		BC: BC{
			B: B{
				B: "b",
			},
			C: C{
				C: "c",
			},
		},
	}

	buf := new(bytes.Buffer)
	if err := toml.NewEncoder(buf).Encode(foo); err != nil {
		log.Fatal(err)
	}

	fmt.Println(buf.String())

	var blob = `
A = "aa"
B = "bb"
C = "cc"
`

	foo2 := &Foo{}
	md, err := toml.Decode(blob, foo2)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Printf("%+v\n", foo2)
	fmt.Println("undecoded:", md.Undecoded())
}
```

This program outputs:

```
A = "a"
A = "a"

[B]
  B = "b"

&{A:{A:aa} BC:{B:{B:bb} C:{C:cc}}}
undecoded: []
```

The TOML duplicates field `A`, does not treat `B` as a top-level field and is missing field `C` altogether. I would have expected the output of `Encode` to look like:

```
A = "a"
B = "b"
C = "c"
```

In the latter part of the example code, you can see that the `blob` variable decodes as expected.

Am I doing something wrong?

curt-hash:
Seems to fix #250

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -319,14 +319,14 @@ func (enc *Encoder) eStruct(key Key, rv reflect.Value) {
 					// tag names as though they are not
 					// anonymous, like encoding/json does.
 					if getOptions(f.Tag).name == "" {
-						addFields(t, frv, f.Index)
+						addFields(t, frv, append(start, f.Index...))
 						continue
 					}
 				case reflect.Ptr:
 					if t.Elem().Kind() == reflect.Struct &&
 						getOptions(f.Tag).name == "" {
 						if !frv.IsNil() {
-							addFields(t.Elem(), frv.Elem(), f.Index)
+							addFields(t.Elem(), frv.Elem(), append(start, f.Index...))
 						}
 						continue
 					}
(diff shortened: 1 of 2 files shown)


arp242:
Thanks!

> I'm not sure if it breaks something else though...

As near as I can tell, this shouldn't break anything else; or at least not that I can think of and encoding/json does something similar.

PR #256, opened 2020-01-17
support parse time.Duration

fengxuway:
Usually, a time.Duration in string format decode to a struct, we must do more things:

First, create time.Duration parse struct:

```go
import (
	"time"
)

type Duration struct {
	time.Duration
}

func (d *Duration) UnmarshalText(text []byte) error {
	var err error
	d.Duration, err = time.ParseDuration(string(text))
	return err
}
```
Then change config struct field for time.Duration. `time.Duration => my.Duration`.

This will cause we must pre process config field for adapt some third library.

So create this merge request for parse time.Duration in string format. And then we don't need to create personal Duration.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -422,6 +422,15 @@ func (md *MetaData) unifyInt(data interface{}, rv reflect.Value) error {
 			panic("unreachable")
 		}
 		return nil
+	} else if d, ok := data.(string); ok {
+		if rv.Type().String() == "time.Duration" {
+			duration, err := time.ParseDuration(d)
+			if err != nil {
+				return e("value %s is invalid format for time.Duration")
+			}
+			rv.SetInt(int64(duration))
+		}
+		return nil
 	}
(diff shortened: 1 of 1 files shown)


dxvgef:
It is necessary !

arp242:
My concern is that the standard library contains many types that could potentially be added: `regexp.RegExp`, `url.URL`, `http.Header`, `mail.Address`, and a bunch more. Should support for all of those added? Unlike datetimes, there is no "native" support for durations in TOML.

On the other hand, if it solves real practical problems that would be hard or impossible to solve otherwise then it can be considered, but it's not clear to me what practical problem this solves now; what do you mean with "pre process config field for adapt some third library"? 

Either way, at the very least this would need to have a test before it can be merged.

arp242:
Closing for now due to lack of feedback. Feel free to get back later and we can reopen. Thanks!

PR #261, opened 2020-05-23
added support and tests for hex and binary ints

iansmith:
I have no idea if this is the "right" way to do it, but it seems to pass the tests.  If you want to advise a better way to do it, I'm happy to do that.


arp242:
I'm closing this one in favour of #260; I still need to review/merge that, but I'll do that soon and closing it now clears the PR list a bit 😅

PR #264, opened 2020-08-05
Support for getting more info (line number and last key) from error when toml.Decode()

link-duan:
Support for getting more info (line number and last key) from error when toml.Decode().
example:

```go
var tomlObject interface{}
content := "..." // toml content
_, err  := toml.Decode(content, &tomlObject)
if err != nil {
    if parseErr, ok := err.(toml.ParseError); ok {
         fmt.Printf("line number: %d, last key is %s", parseErr.Line, parseErr.LastKey)
    }
}
```

### DIFF
--- a/parse.go
+++ b/parse.go
@@ -30,17 +30,22 @@ type parser struct {
 	implicits map[string]bool
 }
 
-type parseError string
+type ParseError struct {
+	Message string
+	Line    int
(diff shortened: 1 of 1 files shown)


arp242:
Thanks! I applied this via #278, where I added your commit and a further one to add a test.

PR #273, opened 2021-06-08
Test recent Go versions in Travis, remove unmaintained notice

arp242:
The tests fail because the https://github.com/BurntSushi/toml-test had a
few testcases added – I'll fix that in a future PR.

Lowest version is Go 1.11 for now as that's what Debian 10/stable uses.
Might bump this to a more recent version later if needed, as that's
really quite old and e.g. the new errors in Go 1.13 might be useful.

Also don't test intermediate Go versions, as with the whole Travis "open
source credits" stuff you run out of them pretty fast and I don't feel
like mucking about with GitHub actions right now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,12 @@
 language: go
 go:
-  - 1.1
-  - 1.2
-  - 1.3
-  - 1.4
-  - 1.5
-  - 1.6
-  - tip
+  - '1.11'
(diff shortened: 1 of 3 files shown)


arp242:
Oh, it looks like Travis isn't on this account any more? Ah well, I guess I'll set up GitHub Actions later.

PR #296, opened 2021-06-19
Download toml-test when use make

sc0Vu:
Hi,

Thanks for this great library. It works great. However, I tried to execute `make` but it failed with toml-text command not found.

In this PR, I download `toml-test` before execute `@${TOML_TEST}` in Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,8 @@ all:
 	@e=0  # So it won't stop on the first command that fails.
 	@go install ./...
 	@go test ./... || e=1
+	@go get github.com/BurntSushi/toml-test/cmd/toml-test@master
 	@${TOML_TEST} -skip="${SKIP_DECODE}" toml-test-decoder || e=1
 	@${TOML_TEST} -encoder -skip="${SKIP_ENCODE}" toml-test-encoder || e=1
+	@go mod tidy
 	@exit $e


arp242:
Thanks! I actually have a better solution for this in the pipeline: with https://github.com/BurntSushi/toml-test/pull/72 the `toml-test` is split out in the CLI and library, and with that I can add:

```
package toml_test

import (
        "testing"

        tomltest "github.com/BurntSushi/toml-test"
)

func TestTomlTest(t *testing.T) {
        r := tomltest.Runner{
                Files:     tomltest.EmbeddedTests(),
                Encoder:   false,
                ParserCmd: []string{"toml-test-decoder"},
        }

        tests, err := r.Run()
        if err != nil {
                t.Fatal(err)
        }

        // TODO: add a String() or Report() method?
        for _, test := range tests.Tests {
                if test.Failed() {
                        t.Errorf("%s\n", test.Failure)
                }
        }
}
```

So you don't need any `toml-test` binary at all and just `go test` works. The Makefile can then just be removed.

The part I still want to fix is to also remove the dependency on the `toml-test-decoder` and `toml-test-encoder` binaries, so you don't need those either (i.e. [this](https://github.com/BurntSushi/toml-test/blob/master/runner.go#L241) should be abstracted a bit more).

You also don't really want to get the latest master version, as that might not work with the current version of this binary. This was a problem in the past, and why the CI uses [the 1.0-beta1 binary](https://github.com/BurntSushi/toml/blob/master/.github/workflows/test.yml#L24). It was actually pretty confusing for people 😅

Since this needs some modification and a better solution will be added this week or so, I think I'll just close this. If you're keen you can help out with a few remaining bits to get that sorted; but that's a bit more work than this PR 😅

arp242:
Added in #297

Issue #60, opened 2014-09-09
Cannot decode integer value to float64 type.


bmatsuo:
It seems like a struct containing a float64 field cannot be decoded from a TOML document containing an integer value. The following example demonstrates this.

```go
package main

import (
	"log"

	"github.com/BurntSushi/toml"
)

func main() {
	raw := "Foo = 123"
	var s struct {
		Foo float64
	}
	_, err := toml.Decode(raw, &s)
	if err != nil {
		log.Fatal(err)
	}
	log.Print(s)
}
```

The error message printed looks like this

> 2014/09/09 13:42:51 Type mismatch for 'struct { Foo float64 }.Foo': Expected float but found 'int64'.

cespare:
I think the behavior is fine as-is. TOML and Go both distinguish between integer and floating-point types; I think it makes sense to enforce that distinction when decoding.

> But seriously, this lossless conversion between numeric types is the natural action in my mind.

It is not a lossless conversion. For example, 10000000000000001 is a valid TOML integer, and a valid Go int64, but cannot be precisely represented as a 64-bit float.

I'll see about improving the error message.

bmatsuo:
@cespare, indeed the spec allows for numbers to exceed the capacity of destination type. It claims (as of this date) to "expect" values within a range/precision. But that is quite different from saying they are invalid. Clearly problems with integer ranges exist with deserialization into other types like int32 and uint. All I would propose is that a conversion to float64 or float32 be handled the same way.

Apparently no other users think as I do given how long this issue has gone on without another supporting comment. So I guess you can do what you will with it. But I still think the package's behavior is incorrect.

If a user is trying to use the Foobaz command with a TOML config file they shouldn't have to wonder if Foobaz is implemented in a language with static types or not when setting numerical configuration values. I think it is just bad UX.

But like I said, if no one agrees then you can close this issue if you want.

cespare:
I don't think either way is "correct" or "incorrect". The TOML spec doesn't give any precise rules about how TOML types have to map to language types. In Go it's natural for TOML integers to map to Go integer types and for TOML floats to map to Go floating-point types.

You're asking for a relaxation which makes certain (uncommon?) code more convenient.

I'm not really opposed to the change, but I don't see this as a failing or bug of the library. If you'd like to send along a PR, I'd be happy to review it. Otherwise, I'll see about improving the error message myself.

cespare:
The error message is now

    toml: cannot decode TOML float into Go value of type int64

[...]

PR #325, opened 2021-11-10
Allow lossless integer to float conversions

saracen:
According to the specification, 64-bit signed integers should be accepted providing they can be represented losslessly.

At the moment, integers to be converted to a float immediately return an error. This change permits integers and only returns an error if they would overflow the float type they're being converted to as they have no problem being represented losslessly.

In addition, this change returns an error if a float is provided but cannot fit within the float type specified by the struct.

This relates to issue https://github.com/BurntSushi/toml/issues/60, which I believe was incorrectly closed:

> BurntSushi:
> This appears to be working as intended. In TOML, floats are indicated by the presence of a decimal point and at least one trailing 0.

Whilst this statement is true, that concerns the input of a float. I believe an argument can be made that the input here concerns an integer, that just so happens to be unmarsharled into a float. It *can* be represented losslessly. Only integers that cannot be should return an error.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -39,9 +39,8 @@ type Primitive struct {
 	context   Key
 }
 
-// The significand precision for float32 and float64 is 24 and
-// 53 bits respectfully. 2^prec-1 gives us the range an integer
-// can be stored within a float without loss of data.
+// The significand precision for float32 and float64 is 24 and 53 bits; this is
+// the range a natural number can be stored in a float without loss of data.
 const (
 	maxSafeFloat32Int = 16777215         // 2^24-1
 	maxSafeFloat64Int = 9007199254740991 // 2^53-1
@@ -225,9 +224,7 @@ func (md *MetaData) unify(data interface{}, rv reflect.Value) error {
 			return e("unsupported type %s", rv.Type())
 		}
 		return md.unifyAnything(data, rv)
-	case reflect.Float32:
-		fallthrough
-	case reflect.Float64:
+	case reflect.Float32, reflect.Float64:
 		return md.unifyFloat64(data, rv)
 	}
 	return e("unsupported type %s", rv.Kind())
--- a/decode.go
+++ b/decode.go
@@ -39,6 +39,14 @@ type Primitive struct {
 	context   Key
 }
 
+// The significand precision for float32 and float64 is 24 and
+// 53 bits respectfully. 2^prec-1 gives us the range an integer
+// can be stored within a float without loss of data.
+const (
+	maxSafeFloat32Int = 16777215         // 2^24-1
+	maxSafeFloat64Int = 9007199254740991 // 2^53-1
+)
+
 // PrimitiveDecode is just like the other `Decode*` functions, except it
 // decodes a TOML value that has already been parsed. Valid primitive values
 // can *only* be obtained from values filled by the decoder functions,
@@ -357,6 +365,9 @@ func (md *MetaData) unifyFloat64(data interface{}, rv reflect.Value) error {
 	if num, ok := data.(float64); ok {
 		switch rv.Kind() {
 		case reflect.Float32:
+			if num < -math.MaxFloat32 || num > math.MaxFloat32 {
+				return e("value %f is out of range for float32", num)
+			}
 			fallthrough
 		case reflect.Float64:
 			rv.SetFloat(num)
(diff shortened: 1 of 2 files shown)


arp242:
Thanks for your patch! I need to think a little bit about this change, read through the other issue, and consider things. I'm feeling pretty sick at the moment so give me a week or so. Feel free to ping me if I forget after this.

It would be helpful to add a little bit of detail on what problem this solves. As in, what is currently hard or impossible to do and how this solves it. Unless I missed it, that's not really mentioned in the issue you linked.

I don't think this is really a "TOML specification issue" as such by the way; it's really up to the implementation to decide what can be accepted for which values depending on what makes sense for the specific language, environment, etc.

saracen:
> It would be helpful to add a little bit of detail on what problem this solves. As in, what is currently hard or impossible to do and how this solves it. Unless I missed it, that's not really mentioned in the issue you linked.

```golang
type Config struct {
    IdleScaleFactor float64 
}
```

If a user inputs `IdleScaleFactor = 1` they receive the error: `toml: cannot load TOML value of type int64 into a Go float`.

`IdleScaleFactor = 1.0` will however work.

One possible solution to this would be to define a `IntOrFloat` struct with an implemented `UnmarshalTOML`. This approach does work, but for us it's slightly more involved to implement because we have other unmarshalers and would need to further implement the same solution for them too.

Our users are typically of a technical background, so it is likely they'll be able to figure out what the problem is from the error message, but it's a poor experience nevertheless.

The arguments in https://github.com/BurntSushi/toml/issues/60 are the same: For their users, it's unintuitive.

So the problem this solves is a user experience issue for anybody populating config and for developers it lifts the burden 
from them needing to implement a special type.

The argument against supporting this was that an `int` to `float` conversion can lose information and that it was a natural in Go for TOML integers to map to Go integer types and TOML floats to map to Go float types.

> I don't think this is really a "TOML specification issue" as such by the way; it's really up to the implementation to decide what can be accepted for which values depending on what makes sense for the specific language, environment, etc.

I think this slightly touches on both a specification issue and what makes sense for the language.

For Go and other libraries, it's not uncommon to expect an integer to be converted to a float upon being unmarshalled:
- The `encoding/json` package can do it: https://play.golang.org/p/_Ajc2z_Npyz
- The `encoding/xml` library can do it: https://play.golang.org/p/K2FmPiAicA9
- The `flag` package can do it: https://play.golang.org/p/P6rXPM_v2bW

Where this becomes a spec issue is that those other decoders convert the provided integer and can lose information, but the TOML specification makes it clear that an error should be thrown if an integer is provided and information would be lost.

For other conversions, this library already adheres to that rule. A really large number would not fit within a `uint8` and an error is thrown. This PR does just that with integer to float conversions, ensuring that if information would be lost, an error is thrown.

> I'm feeling pretty sick at the moment so give me a week or so. Feel free to ping me if I forget after this.

I'm sorry to hear that, I hope you feel better soon! There's no rush. I'll catch up with you once you're better. Thank you!

arp242:
I rebased this branch on master and added a commit with some boring/pedantic stylistic changes I didn't want to bother you with.

I think this mostly looks okay; but I'm not sure we need to check for maximum safe natural numbers; if I specified a float in Go then I'm fine with some (possible) loss of accuracy, right? If you want/need fully accurate representation then you should use an int. It's useful to have this check in some use cases, and probably unwanted in other use cases; if you want such a check, then it's something people should implement themselves.

arp242:
e.g. the encoding/json library also doesn't check for this: https://play.golang.org/p/faOdIf91u-J

That second value is inaccurate, but that's just how floats work. I think that check should probably be removed. What do you think?

saracen:
@arp242 

I think the float32 boundary check could be removed. I'd expect loss here, and unlikely integers, the TOML spec makes no mention about them being represented losslessly.

The error check for int -> float I'd keep, but only because the spec says:
> If an integer cannot be represented losslessly, an error must be thrown.

In practice, you could probably remove both checks and just cast to float with loss in both cases and nobody would complain 👍 I don't mind which route we take.

arp242:
Oh, right. Well, let's keep it then. If it really proves to be a problem for some people later on we can always relax things (possibly via an option), which will be a compatible change. The reverse of "don't add check and add one later" won't be compatible as it has a chance of breaking people's existing TOML files.

Issue #169, opened 2017-03-24
How to use IsDefined with array of tables?


gorankor:
I don't know if this is an issue or just me not knowing how to properly call it, so I apologize if I opened a new issue which is not an issue at all.

I have a config file where I want to specify one section several times:

> [[methods]]
> foo = "method1"
> namespace = "ns1"
> 
> [[methods]]
> foo = "method2"
> namespace = "ns2"`
> 
> ...

Here's a complete example:

```
package main

import (
        "fmt"
        "os"
        "github.com/BurntSushi/toml"
)

type confMethod struct {
    Foo         string     `toml:"foo,omitempty"`
    NameSpace   string     `toml:"namespace,omitempty"`
}

// Methods struct - used in include files
type confMethods struct {
    Methods []confMethod     `toml:"methods,omitempty"`
}

func main() {
    var mConf confMethods

    c := `[[methods]]
foo = "method1"
namespace = "root/cimv2"

[[methods]]
foo = "method2"
namespace = "root/cimv2"`

    md, err := toml.Decode(c, &mConf)
    if err != nil {
        fmt.Fprintf(os.Stderr, "Error: %s\n", err)
        os.Exit(1)
    }

    fmt.Printf("keys=%s\n", md.Keys())

    if md.IsDefined("methods", "namespace") {
        fmt.Printf("defined\n")
    } else {
        fmt.Printf("not defined\n")
    }
}
```

This returns "not defined".  How can one use IsDefined() in this case? Is it possible at all? 

I have tried different usages as the commend above IsDefined() in decode.go says:

```
md.IsDefined("methods", "namespace")
md.IsDefined("methods", "methods.namespace")
md.IsDefined("methods", "method", "namespace")
md.IsDefined("namespace")
```
but nothing works.

My real example is a bit more complex, and config parser works as expected when I just blindly assign values from config struct, but I'd like to only do that when something is present in the config file. I can't rely on what Keys() returns as that one is not specific enough, but it at least shows that the keys are there as expected (as the example prints it).

If I look at what is going on in decode.go and compare working (where I use just a simple struct) vs. non-working example, when it works, it gets this type:

```map[string]interface {}```

But for this failing example it gets:

```[]map[string]interface {}```

And then it fails here on second (might be third - depends how it gets called obviouslly) loop iteration:

```if hash, ok = hashOrVal.(map[string]interface{}); !ok```

cespare:
[...]

PR #328, opened 2021-11-16
Support encoding comments and specifying the encoding format

### LINKED ISSUE #64: Any support for encoding strings as multi-line?

### LINKED ISSUE #75: Support comments in encoding

### LINKED ISSUE #160: Support encoding inline tables(?)
See comments in #70.

I think that if we had #76, though, we might be able to allow people who want to to emit their own inline tables.

I'm loath to add more special struct tags or whatnot.

### LINKED ISSUE #192: proposal: re-use existing map items during decode map
I have map, that has structures as elements.

```go
type LogConfig struct {
     Filename string  `toml:"file"`
     Level       string  `toml:"level"`
}

type Config struct {
    Loggers map[string]*LogConfig `toml:"log"`
}

fun NewConfig() *Config {
    return &Config{
         Loggers: map[string]*LogConfig {
                 "error": &LogConfig{Level: "error"},
                 "access": &LogConfig{Level: "info"},
        }
}

cfg := NewConfig()
toml.DecodeFile("config.toml", &cfg)
```

And If I do not specify the both "level and file" in config, I lose the defaults with I specify on creating `Config` because `tome.Decode` always re-creates map items during decode.

As user I expect that if some option is not specified in config, it will keep original value.

PS:
I agree that encoding/json has the similar behaviour, https://play.golang.org/p/d6t50gCppF
but it is not useful for configs.

I found the following workarounds:
 - use structure instead of map
 - re-init LogConfig after decode.

but they are not convenient and makes code more complex.

### LINKED ISSUE #213: support round tripping with comments
This is related to #75 but the solution might look quite different.

TOML is being used for the dep tool, and a major obstacle to making that tool easier to use is the fact that the Gopkg.toml file cannot be programmatically manipulated without loss of user-edited content.

If the toml package supported round-tripping the TOML while retaining comments, I believe that it would help a great deal.

I'm afraid I'm sure how the API would look though. Some degree of loss is probably inevitable, but it may be possible to a job that's good enough for all practical cases.

### LINKED ISSUE #269: Use the Encode function to Encode structures as "" (multi-line literal strings wrapped in three single quotes)
golang 
type Item struct {
	Type  RedisKeyType `toml:"type"`
	Key   string       `toml:"key"`
	Value interface{}  `toml:"value"`
}
encode-------
toml 
type="string"
key="test"
value=''' 
{}
'''

arp242:
This allows encoding comments and setting some flags to control the
format. While toml.Marshaler added in #327 gives full control over how
you want to format something, I don't think it's especially
user-friendly to tell everyone to create a new type with all the
appropriate formatting, escaping, etc. The vast majority of of use cases
probably just call in to a few simple categories such as "use `"""` or
"encode this number as a hex".

I grouped both features together as they're closely related: both set
additional information on how to write keys.

What I want is something that:

1. allows setting attributes programmatically;
2. supports round-tripping by default on a standard struct;
3. plays well with other encoders/decoders;
4. has a reasonable uncumbersome API.

Most options (custom types, struct tags) fail at least one of these;
there were some PRs for struct tags, but they fail at 1, 2, and
(arguably) 4. Custom types fail at 2, 3, and probably 4.

---

This adds SetMeta() to the Encoder type; this is already what we have
when decoding, and all additional information will be set on it. On
MetaData we add the following types:

    SetType()       Set TOML type info.
    TypeInfo()      Get TOML type info.
    Doc()           Set "doc comment" above the key.
    Comment()       Set "inline comment" after the key.

Every TOML type has a type in this package, which support different
formatting options (see type_toml.go):

    Bool
    String
    Int
    Float
    Datetime
    Table
    Array
    ArrayTable

For example:

    meta := toml.NewMetaData().
            SetType("key", toml.Int{Width: 4, Base: 16}).
            Doc("key", "A codepoint").
    Comment("key", "ë")
    toml.NewEncoder(os.Stdout).SetMeta(meta).Encode(struct {
            Key string `toml:"key"`
    }{"ë")

Would write:

    # A codepoint.
    key = 0x00eb  # ë

It also has Key() to set both:

    toml.NewMetaData().
        Key("key", toml.Int{Width: 4, Base: 16}, toml.Doc("A codepoint"), toml.Comment("ë")).
        Key("other", toml.Comment("..."))

The advantage of this is that it reduces the number of times you have to
type the key string to 1, but it uses interface{}. Not yet decided which
one I'll stick with, and also not a huge fan of Doc() and Comment(), but
I can't really think of anything clearer at the moment (these are the
names the Go ast uses).

---

The Decode() sets all this information on the MetaData, so this:

    meta, _ := toml.Decode(..)
    toml.NewEncoder(os.Stdout).SetMeta(meta).Encode(..)

Will write it out as "key = 0x00eb" again, rather than "key = 235".

This way, pretty much any flag can be added programmatically without
getting in the way of JSON/YAML/whatnot encoding/decoding.

---

I don't especially care how you need to pass the keys as strings, but
there isn't really any good way to do it otherwise. There is also the
problem that the "key" as found in the parser may be different than the
"key" the user is expecting if you don't use toml struct tags:

    type X struct { Key int }

Will read "key = 2" in to "Key", but when encoding it will write as
"Key" rather than "key". The type information will be set to "key", but
when encoding it will look for "Key", so round-tripping won't work
correct and has the potential for confusion if the wrong key is set.

This is not so easy to fix since we don't have access to the struct in
the parser. I think it's fine to just document this as a caveat and tell
people to use struct tags, which is a good idea in any case.

---

I'm not necessarily opposed to also adding struct tags for most of these
things, although I'm not a huge fan of them. Since struct tags can't be
set programmatically it's not really suitable for many use cases (e.g.
setting comments dynamically, using multiline strings only if the string
contains newlines, etc.) It's something that could maybe be added in a
future PR, if a lot of people ask for it.

Fixes #64
Fixes #75
Fixes #160
Fixes #192
Fixes #213
Fixes #269

### DIFF
--- a/cmd/toml-test-decoder/main.go
+++ b/cmd/toml-test-decoder/main.go
@@ -31,13 +31,14 @@ func main() {
 	}
 
 	var decoded interface{}
-	if _, err := toml.DecodeReader(os.Stdin, &decoded); err != nil {
+	meta, err := toml.DecodeReader(os.Stdin, &decoded)
+	if err != nil {
 		log.Fatalf("Error decoding TOML: %s", err)
 	}
 
 	j := json.NewEncoder(os.Stdout)
 	j.SetIndent("", "  ")
-	if err := j.Encode(tag.Add("", decoded)); err != nil {
+	if err := j.Encode(tag.Add(meta, "", decoded)); err != nil {
 		log.Fatalf("Error encoding JSON: %s", err)
 	}
 }
--- a/decode.go
+++ b/decode.go
@@ -128,11 +128,12 @@ func (dec *Decoder) Decode(v interface{}) (MetaData, error) {
 		return MetaData{}, err
 	}
 	md := MetaData{
-		mapping: p.mapping,
-		types:   p.types,
-		keys:    p.ordered,
-		decoded: make(map[string]bool, len(p.ordered)),
-		context: nil,
+		mapping:  p.mapping,
+		types:    p.types,
+		keys:     p.ordered,
+		comments: p.comments,
+		decoded:  make(map[string]bool, len(p.ordered)),
+		context:  nil,
 	}
 	return md, md.unify(p.mapping, indirect(rv))
 }
@@ -462,6 +463,7 @@ func (md *MetaData) unifyText(data interface{}, v encoding.TextUnmarshaler) erro
 	var s string
 	switch sdata := data.(type) {
 	case Marshaler:
+		fmt.Println("unifyText (Marshaler)", data, "in to", v)
 		text, err := sdata.MarshalTOML()
 		if err != nil {
 			return err
@@ -473,6 +475,7 @@ func (md *MetaData) unifyText(data interface{}, v encoding.TextUnmarshaler) erro
 			return err
 		}
 		s = string(text)
+		// fmt.Println("unifyText (TextMarshaler)", data, "in to", v, "=", s)
 	case fmt.Stringer:
 		s = sdata.String()
 	case string:
--- a/encode.go
+++ b/encode.go
@@ -2,6 +2,7 @@ package toml
 
 import (
 	"bufio"
+	"bytes"
 	"encoding"
 	"errors"
 	"fmt"
@@ -98,8 +99,16 @@ type Encoder struct {
 	// String to use for a single indentation level; default is two spaces.
 	Indent string
 
+	// TODO(v2): Ident should be a function so we can do:
+	//
+	//   NewEncoder(os.Stdout).SetIndent("prefix", "indent").MetaData(meta).Encode()
+	//
+	// Prefix is also useful to have.
+
 	w          *bufio.Writer
 	hasWritten bool // written any output to w yet?
+	wroteNL    int  // How many newlines do we have in a row?
+	meta       *MetaData
 }
 
 // NewEncoder create a new Encoder.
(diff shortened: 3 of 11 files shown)


tschmidtb51:
@arp242 Any updates on this one?

arp242:
> Any updates on this one?

I ran in to a problem: to support this properly I need to change the way keys are addressed in Meta type, because right now you can't address certain keys: see #169.

So that syntax needs to be changed, but isn't a backwards-compatible change :-( My plan is to fix most of the outstanding bugs etc. and then release a v2 with this change, but I just haven't had the time.

tschmidtb51:
Thanks for the update!

PR #337, opened 2021-12-25
Setup golangci-lint and code coverage 

mmorel-35:
This provides two new steps in the CI workflow.
The first is golangci-lint.
The other is the report of the code coverage withe codecov.io.
There is also the setup of dependabot to keep github-actions dependencies up-to-date


arp242:
Thanks for your work!

I'm not a huge fan of these expansive CI linter setups; I find it's a lot of kerfuffle with minimal benefits, as well as some drawbacks. None of the changes here are really all that meaningful: `var foo = ` vs. `foo :=` or a newline more or less: I don't think it really matters, and I don't want to bother people submitting patches with pedantic stuff like that, even if it's from a CI tool.

It can still be a good trade-off for some projects, especially with large teams (although even in those cases not necessarily), but for something like this I don't think it's a very good trade-off, so I don't think I'll merge this.

PR #339, opened 2022-01-12
Decode S=[] into a non-nil []interface{}.

### LINKED ISSUE #338: x = [] decodes into nil
The following program works differently between 0.3.1 and 0.4.1:

```
package main

import (
    "log"
    "reflect"

    "github.com/BurntSushi/toml"
)

func main() {
    blob := `
    x = []
    `
    var d map[string]interface{}
    if _, err := toml.Decode(blob, &d); err != nil {
        log.Fatal(err)
    }

    x := d["x"]
    log.Println(x, reflect.ValueOf(x).IsNil())
}
```

On 0.3.1,
```
[] false
```

On 0.4.1,
```
[] true
```

For my use case, this is a problem because I'm subsequently encoding the value into JSON and it writes `x = null` instead of the expected `x = []`. It also goes deeper than that, for instance `x = [[]]` would get JSONed as `x = [null]`.

Is this an intended change? Do you think it would be possible/correct to restore the previous behaviour?

dtcaciuc:
This restores v0.3 behaviour and also makes it
consistent with how the same decodes into something
like struct { S []string }.

Closes #338

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -618,6 +618,17 @@ func TestDecodeSlices(t *testing.T) {
 	}
 }
 
+func TestDecodeInterfaceSlice(t *testing.T) {
+	var v map[string]interface{}
+	if _, err := Decode(`S = []`, &v); err != nil {
+		t.Errorf(err.Error())
+	}
+	// S = [] should decode into a non-nil []interface{} (see issue #338).
+	if reflect.ValueOf(v["S"]).IsNil() {
+		t.Errorf("S is nil")
+	}
+}
+
 func TestDecodePrimitive(t *testing.T) {
 	type S struct {
 		P Primitive
--- a/parse.go
+++ b/parse.go
@@ -350,6 +350,10 @@ func (p *parser) valueArray(it item) (interface{}, tomlType) {
 		array []interface{}
 		types []tomlType
 	)
+	// Initialize to a non-nil empty slice. This makes it
+	// consistent with how S = [] decodes into a non-nil slice
+	// inside something like struct { S []string }.
+	array = []interface{}{}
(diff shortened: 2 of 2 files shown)


dtcaciuc:
The struct behaviour I'm referring to is already tested in [TestDecodeSlices](https://github.com/BurntSushi/toml/blob/7d0236fe74765a731cf768bea23b9132b8c2b9e0/decode_test.go#L605)

decode_test.go
```diff
@@ -618,6 +618,17 @@ func TestDecodeSlices(t *testing.T) {
 	}
 }
 
+func TestDecodeInterfaceSlice(t *testing.T) {
+	var v map[string]interface{}
+	if _, err := Decode(`S = []`, &v); err != nil {
```
arp242, line 623: Could you also test nested arrays and inline tables?

```
arr = []
nested = [[]]
tbl = {arr = []}
```

Just to make sure this keeps working in all future updates!

parse.go
```diff
@@ -350,6 +350,10 @@ func (p *parser) valueArray(it item) (interface{}, tomlType) {
 		array []interface{}
 		types []tomlType
 	)
+	// Initialize to a non-nil empty slice. This makes it
+	// consistent with how S = [] decodes into a non-nil slice
+	// inside something like struct { S []string }.
+	array = []interface{}{}
```
arp242, line 356: Could you mention your issue in this comment? Just for future reference.

Also, might as well move this up with the declaration: `array = []interface{}{}` rather than defining it and then assigning a value.

decode_test.go
arp242: Actually, looking at this a bit closer the tests should be integrated with `TestDecodeSlices`, I'll just merge this as-is and fix it up; thanks!

Issue #341, opened 2022-01-24
Marshaler Should not writeQuoted string


lucasbutn:
Hi, hope you can understand this issue (if it is really an issue).

In enconde.go Line 210
case Marshaler:
		s, err := v.MarshalTOML()
		if err != nil {
			encPanic(err)
		}
		enc.writeQuoted(string(s))
		return

If the struct implements MarshalToml to output a valid toml, this piece of code is writing the result with "double quotes" which is the same as TextMarshaler, which I don't think this is meant to this.

What I'm trying to do is to Marshall an structure to a single value, and I'm getting a quoted value

In my case: 

If I have:
type WrappedBool struct {
    Value bool
}

And I what to marshall this:

type SomeStruct struct {
       BooleanValue *WrappedBool
}

Without custom marshall:

[SomeStruct]
    [SomeStruct.BooleanValue] = true
    
With Custom Marshall:

[SomeStruct]
    SomeValue = "true"
    
What it actually should be

[SomeStruct]
    SomeValue = true

This double quotes are added outside the MarshalTOML() function and according to the interface comments, it should not be like this right?:

// Marshaler is the interface implemented by types that can marshal themselves
// into valid TOML.
type Marshaler interface {
	MarshalTOML() ([]byte, error)
}

Or this is some expected behavior? 

Is there is another way to implement what I'm trying to do and I didn't figure it out?

arp242:
I think this is because I copied this by accident from the already existing TextMarshaler. You should be able to write `true` rather than `"true"` in any case. I'll have a more detailed look later.

lucasbutn:
> I think this is because I copied this by accident from the already existing TextMarshaler. You should be able to write `true` rather than `"true"` in any case. I'll have a more detailed look later.

Hi, did you had any chance to took at this issue. I'll share here some code to test it

https://go.dev/play/p/CkqUjfEuZlK

See, I've compared to json marshal to illustrate the expected behaviour

Just run it and you will see what is wrong. MarshalTOML result should no be wrote with quotes. But please confirm

arp242:
> did you had any chance to took at this issue.

No, sorry;  I haven't had the time. Not sure if I have in the coming weeks, but I'll be happy to review and merge patches in a timely fashion.

PR #344, opened 2022-02-09
Removed write quote in marshal to allow write other types than strings

lucasbutn:
Main issue was that in enconde.go Line 215, the marshaled result was wrote with quotes, making imposible to write other values than strings.

Now the implementer of marshal, is responsible to write the quotes if required, as it happens in other marshalers if we compare it to JSON marshal, but it gives more flexibility.

The full issue is explained here: https://github.com/BurntSushi/toml/issues/341

An easy way to verify the issue is change enconde.go Line 215 to its previous state an run the test


arp242:
The existing `TestEncodeTOMLMarshaler` seems to fail now; I think fixing that is just a matter of mucking about with quotes in the `MarshalTOML()` methods, since it doesn't add any now. I fixed the CI in master, so if you rebase that should work.

But other than that, looks good.

lucasbutn:
> The existing `TestEncodeTOMLMarshaler` seems to fail now; I think fixing that is just a matter of mucking about with quotes in the `MarshalTOML()` methods, since it doesn't add any now. I fixed the CI in master, so if you rebase that should work.
> 
> But other than that, looks good.

Sorry, I've neglect the tests. I've fixed them and add extra coverage for Slices and Complex.

arp242:
Not sure why I needed to approve the CI runs again 🤔 Anyway, now the Go 1.13 tests fails with:

    Error: ./custom_marshaler_test.go:170:23: undefined: strconv.FormatComplex
    Error: ./custom_marshaler_test.go:178:17: undefined: strconv.ParseComplex

Looks like these functions were added in Go 1.15, whereas this library tries to be compatible with Go 1.13 and newer (for the new errors interface); I'm not opposed to bumping that to a newer version, but since it's just for a test I'd rather not since you can get the same effect with some basic string parsing. Bit annoying it's extra work though; sorry about that :-(

lucasbutn:
> Not sure why I needed to approve the CI runs again 🤔 Anyway, now the Go 1.13 tests fails with:
> 
> ```
> Error: ./custom_marshaler_test.go:170:23: undefined: strconv.FormatComplex
> Error: ./custom_marshaler_test.go:178:17: undefined: strconv.ParseComplex
> ```
> 
> Looks like these functions were added in Go 1.15, whereas this library tries to be compatible with Go 1.13 and newer (for the new errors interface); I'm not opposed to bumping that to a newer version, but since it's just for a test I'd rather not since you can get the same effect with some basic string parsing. Bit annoying it's extra work though; sorry about that :-(

I've just removed the complex testing for now. It is no different from any CustomMarshall to a string format annyway. since toml doest'n suppors '(' and ')', or '+' there is no other way to marshal a complex than a string in this case.
Sorry I didn't had any tima before

Thanks. Hope CI runs ok this time

PR #346, opened 2022-02-18
Support time.Duration string format

husio:
Using numeric representation of time.Duration is confusing, because
value is stored as UNIX nanoseconds. Humans tend to be more comfortable
with the string representation of the time.Duration.

In addition to time.Duration being an int64 number and encoding/decoding
as such, allow to use its string representation. For example, the below
structure:

    Configuration := struct {
    	ExpireIn time.Duration
    }{
    	ExpireIn 15 * time.Minute,
    }

Is currently serialized as:

    ExpireIn = 900000000000

This commit is prioritizing its string representation and the new
encoding will produce:

    ExpireIn = "15m0s"

This change is backward compatible. Both old numeric value and the new
string reprentation are accepted by the decoder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,5 @@
 TOML stands for Tom's Obvious, Minimal Language. This Go package provides a
-reflection interface similar to Go's standard library `json` and `xml`
-packages.
+reflection interface similar to Go's standard library `json` and `xml` packages.
 
 Compatible with TOML version [v1.0.0](https://toml.io/en/v1.0.0).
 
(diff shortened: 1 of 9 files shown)


arp242:
I put some further work in the [durations2 branch](https://github.com/BurntSushi/toml/tree/duration2) btw; turns out I can't force push here after a rebase. Sorry, I kinda hijacked your PR here 😅

Also needs a test for `*time.Duration`, `map[string]time.Duration`, and `map[string]*time.Duration` btw; just mentioning it so it won't get forgotten.

arp242:
Okay, I fixed the error reporting today, and you can now get nice errors:

    toml: error: invalid duration: "99 bottles"

    At line 1, column 5-15:

        1 | D = "99 bottles"
                ^^^^^^^^^^
    Error help:

        A duration must be as "number<unit>", without any spaces. Valid units are:

            ns         nanoseconds (billionth of a second)
            us, µs     microseconds (millionth of a second)
            ms         milliseconds (thousands of a second)
            s          seconds
            m          minutes
            h          hours

        You can combine multiple units; for example "5m10s" for 5 minutes and 10
        seconds.

I merged the `durations2` branch which includes your commit, so this can be "closed as merged".

PR #348, opened 2022-02-23
Test ARM + 32bit

arp242:
So e.g. https://github.com/BurntSushi/toml/pull/347 is caught. Figure I
might as well use ARM instead of i386.


arp242:
Actually, running the tests as ARM doesn't work (obviously...) so just use GOARCH=386.

PR #350, opened 2022-04-07
Update decode.go

FPbear:
Special case. Handle time.Time values specifically.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"reflect"
 	"strings"
+	"time"
 )
 
 // Unmarshaler is the interface implemented by objects that can unmarshal a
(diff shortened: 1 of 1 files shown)


arp242:
It would be helpful if you could describe a little bit what this. Does this fix a bug? Or add a feature? What code didn't work before and works now?

FPbear:
Special case. Handle time.Time values specifically.
Without this processing, when the time is in the value of the map, the data will not be properly parsed and cause panic

arp242:
Ah right, "doesn't work when unmarshaling to a map" is what I was looking for, thanks! I wrote a test and it does indeed seem to fail.

I took a look, and I'm not 100% sure yet if this is the best way to fix it; it should ideally work for all types that implement the TextUnmarshaler interface, and not just `time.Time` (as the comment says, I assume you copied that from an old version?) I'll have to take another deeper look later. Feel free to ping if I forget 😅

kkHAIKE:
i'm sure this change happened in v0.3.1 -> v0.4.0
https://github.com/BurntSushi/toml/compare/v0.3.1...v0.4.0#diff-7f87da95366e3cf56cfcb5904ee3127a3e89e6152a6a004a7e3de529e3a8a937L173-L180
屏幕快照 2022-04-27 下午6 40 09

kkHAIKE:
root cause is map value time.Time not impl TextUnmarshaler(buf *time.Time impled).

so can't trigger this place
https://github.com/BurntSushi/toml/blob/891d2617ddbdfa265c4095b53103c010c98e6591/decode.go#L207-L210

FPbear:
> Ah right, "doesn't work when unmarshaling to a map" is what I was looking for, thanks! I wrote a test and it does indeed seem to fail.
> 
> I took a look, and I'm not 100% sure yet if this is the best way to fix it; it should ideally work for all types that implement the TextUnmarshaler interface, and not just `time.Time` (as the comment says, I assume you copied that from an old version?) I'll have to take another deeper look later. Feel free to ping if I forget 😅

@arp242 
Yes, this is copied from an older version.
Is there a better way to deal with this problem now? 😅

arp242:
I merged #354, which should address this problem.

PR #356, opened 2022-05-29
omitempty also considers structs with all zero values "empty".

arp242:
A common use case for this is time.Time, which will get marshaled
otherwise.

A struct with private fields set is *not* considered to be "empty"; not
sure yet if this is the best behaviour 🤔 It is the easiest to implement
though.

Fixes #353


arp242:
Actually, checking only exported fields doesn't really work; `time.Time` for example contains only unexported fields and everything is done through methods, so it will always be "empty".

I guess the current behaviour is fine.

Issue #357, opened 2022-06-04
Unexpected behaviour with the TextMarshaler interface


alessandro-guido:
The toml packages behaves unexpectedly when using the TextMarshaler interface. In particular, the toml package ignores a MarshalText implementation if the receiver is a pointer but the value to be serialized is not, like in the following example.

```
type Document struct {
	Field IntSlice
}

type IntSlice []int

func (items *IntSlice) MarshalText() ([]byte, error) {
    ...
}
```

I've tried to setup a complete example here: https://go.dev/play/p/tMLzMg5sqTe. The json package, IMHO correctly, makes use of the MarshalText defined as above. Changing the signature like below is required to make the toml package use the custom MarshalText function.
```
func (items IntSlice) MarshalText() ([]byte, error) {
```

Decoding interface is not affected by this problem, i.e. an UnmarshalText function is picked even if it has a pointer type receiver (tested locally, not shown in the example code).

arp242:
Does this still happen in the latest master? Some changes were made last week (specifically: #354).

alessandro-guido:
Yes, it does. Tested with BurntSushi/toml@v1.1.1-0.20220529223624-8fccf3c6afbe

arp242:
Alrighty, just checking!

There is actually already a test case for this: https://github.com/BurntSushi/toml/blob/master/encode_test.go#L342-L343 – although it tests that it "fails" rather than works.

I took a quick look, and encoding/json contains some additional code to handle this, in `newTypeEncoder()`. Need to look a bit at how that works and adjust `Encoder.encoder` accordingly here.

To set expectations: I probably won't have time for this in the next few weeks, but I'll review patches if you feel like taking a crack at it.

kkHAIKE:
@arp242 make a fix for it, https://github.com/BurntSushi/toml/pull/358
😊

PR #358, opened 2022-06-16
change eindirect behave match with indirect from decode

kkHAIKE:
fix https://github.com/BurntSushi/toml/issues/357

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -64,6 +64,12 @@ var dblQuotedReplacer = strings.NewReplacer(
 	"\x7f", `\u007f`,
 )
 
+var (
+	marshalToml = reflect.TypeOf((*Marshaler)(nil)).Elem()
+	marshalText = reflect.TypeOf((*encoding.TextMarshaler)(nil)).Elem()
+	timeType    = reflect.TypeOf((*time.Time)(nil)).Elem()
+)
+
 // Marshaler is the interface implemented by types that can marshal themselves
 // into valid TOML.
 type Marshaler interface {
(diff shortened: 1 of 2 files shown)


arp242:
Thanks!

One thing that's missing is a few tests that test the new behaviour.

kkHAIKE:
add test.
but i found origin TestEncodeTextMarshaler has bug with Sound field. when change Encode obj to '&x', but use my fix is work around....nnn

output:
```toml
Name = "Goblok"

[Sound]
  S = "miauw"
Sound2 = "miauw"
Food = "chicken, fish"
Food2 = "chicken, fish"
Complex = "(42.000000+666.000000i)"
Fun = "why would you do this?"

[Labels]
  color = "black"
  type = "cat"
```

kkHAIKE:
i found other bug with embed private struct/interface, let me make new commit

kkHAIKE:
fixed with privete embed struct / nostruct field, behaviour like json.Marshal

arp242:
> but i found origin TestEncodeTextMarshaler has bug with Sound field. when change Encode obj to '&x', but use my fix is work around

Yeah, that's expected; that's what I mentioned in the issue as well.

Anyway, seems good at a glance, but I'll go over your PR in detail later.

arp242:
Thank you!

kkHAIKE:
Can you create a new tag for last commit 😊

arp242:
I want to pick up a few more small things before a new release @kkHAIKE, because then all non-v2 issues are resolved and I can start working on that. In the meanwhile, you can use a new version in your project by using `go get github.com/BurntSushi/toml@master` (you don't *need* a tagged release to get this patch).

PR #361, opened 2022-07-28
Don't panic with 'omitempty' and uncomparable type

### LINKED ISSUE #360: Encoding: 'omitempty' will panic if the struct contains a uncomparable type
While encoding a struct will be checked for emptiness with the isEmpty function:
```golang
func isEmpty(rv reflect.Value) bool {
	switch rv.Kind() {
	case reflect.Array, reflect.Slice, reflect.Map, reflect.String:
		return rv.Len() == 0
	case reflect.Struct:
		return reflect.Zero(rv.Type()).Interface() == rv.Interface()
	case reflect.Bool:
		return !rv.Bool()
	}
	return false
}
```
The comparing of structs panics if the structs contains any uncomparable type.

Minimal example:
```golang
type a struct {
	B b `toml:"B,omitempty"`
}

type b struct {
	C []string `toml:"C,omitempty"`
}

func TestToml(t *testing.T) {
	b1 := b{C: []string{"foo"}}
	a1 := a{B: b1}
	var buf bytes.Buffer
	toml.NewEncoder(&buf).Encode(a1)
	fmt.Printf("%s", buf.String())
}
```

Output:
```
=== RUN   TestToml
--- FAIL: TestToml (0.00s)
panic: runtime error: comparing uncomparable type data.b [recovered]
	panic: runtime error: comparing uncomparable type data.b [recovered]
	panic: runtime error: comparing uncomparable type data.b
```

arp242:
Fixes #360

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -261,7 +261,7 @@ func (enc *Encoder) eElement(rv reflect.Value) {
 			enc.eElement(reflect.ValueOf(v))
 			return
 		}
-		encPanic(errors.New(fmt.Sprintf("Unable to convert \"%s\" to neither int64 nor float64", n)))
+		encPanic(fmt.Errorf("unable to convert %q to int64 or float64", n))
 	}
 
 	switch rv.Kind() {
@@ -504,7 +504,8 @@ func (enc *Encoder) eStruct(key Key, rv reflect.Value, inline bool) {
 			if opts.name != "" {
 				keyName = opts.name
 			}
-			if opts.omitempty && isEmpty(fieldVal) {
+
+			if opts.omitempty && enc.isEmpty(fieldVal) {
 				continue
 			}
 			if opts.omitzero && isZero(fieldVal) {
@@ -648,11 +649,14 @@ func isZero(rv reflect.Value) bool {
 	return false
 }
 
-func isEmpty(rv reflect.Value) bool {
+func (enc *Encoder) isEmpty(rv reflect.Value) bool {
 	switch rv.Kind() {
 	case reflect.Array, reflect.Slice, reflect.Map, reflect.String:
 		return rv.Len() == 0
 	case reflect.Struct:
+		if !rv.Type().Comparable() {
+			encPanic(fmt.Errorf("type %q cannot be used with omitempty as it's uncomparable", rv.Type()))
+		}
 		return reflect.Zero(rv.Type()).Interface() == rv.Interface()
 	case reflect.Bool:
 		return !rv.Bool()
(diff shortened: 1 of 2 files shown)


kkHAIKE:
i think use Value.IsZero is better..😊

arp242:
> i think use Value.IsZero is better..blush

You mean checking if the struct has the `IsZero() bool` method? I looked at that before in #353, and decided to go with this method. Is there any particular reason why you think IsZero() is better?

kkHAIKE:
than not need use Comparable() to check..😊

arp242:
I had to go back and look why I didn't use it @kkHAIKE, but the problem is that it considers e.g. `[]string{}` to be non-zero, whereas we want to check the length.

Example program:

```
package main

import (
	"fmt"
	"reflect"
)

func main() {
	type Strukt struct{ Slice []int }

	a := reflect.ValueOf(Strukt{Slice: []int{1}})
	b := reflect.ValueOf(Strukt{Slice: []int{}})
	c := reflect.ValueOf(Strukt{})

	fmt.Printf("%-10s %-8s %s\n", "", "isZero", "isEmpty")
	fmt.Printf("%-10s %-8v %v\n", "values", a.IsZero(), isEmpty(a))
	fmt.Printf("%-10s %-8v %v\n", "[]int{}", b.IsZero(), isEmpty(b))
	fmt.Printf("%-10s %-8v %v\n", "no value", c.IsZero(), isEmpty(c))

}

func isEmpty(rv reflect.Value) bool {
	switch rv.Kind() {
	case reflect.Array, reflect.Slice, reflect.Map, reflect.String:
		return rv.Len() == 0
	case reflect.Struct:
		if rv.Type().Comparable() {
			return reflect.Zero(rv.Type()).Interface() == rv.Interface()
		}
		// Not comparable; need to loop over all fields and call isEmpty()
		// recursively so that we check the length for slices etc.
		for i := 0; i < rv.NumField(); i++ {
			if isEmpty(rv.Field(i)) {
				return true
			}
		}
	case reflect.Bool:
		return !rv.Bool()
	}
	return false
}
```

Outputs:

```
           isZero   isEmpty
values     false    false
[]int{}    false    true
no value   true     true
```

kkHAIKE:
oh.. sorry, i forgot this case...change the impl to that
```golang
func isEmpty(rv reflect.Value) bool {
	switch rv.Kind() {
	case reflect.Array, reflect.Slice, reflect.Map:
		return rv.Len() == 0
	case reflect.Struct:
		for i := 0; i < rv.NumField(); i++ {
			if !isEmpty(rv.Field(i)) {
				return false
			}
		}
		return true
	}
	return rv.IsZero()
}
```

arp242:
Cheers, thanks; I mentioned the discussion in #360; might be a few weeks before I look at this, but happy to accept PRs as long as they add decent test cases!

PR #362, opened 2022-08-05
add isEmptyStruct func

kkHAIKE:
related to
https://github.com/BurntSushi/toml/issues/360

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -649,14 +649,34 @@ func isZero(rv reflect.Value) bool {
 	return false
 }
 
+func isEmptyStruct(rv reflect.Value) bool {
+	switch rv.Kind() {
+	case reflect.Array:
+		for i := 0; i < rv.Len(); i++ {
+			if !isEmptyStruct(rv.Index(i)) {
+				return false
+			}
+		}
+		return true
+	case reflect.Slice, reflect.Map:
+		return rv.Len() == 0
+	case reflect.Struct:
+		for i := 0; i < rv.NumField(); i++ {
+			if !isEmptyStruct(rv.Field(i)) {
+				return false
+			}
+		}
+		return true
+	}
+	return rv.IsZero()
+}
+
 func (enc *Encoder) isEmpty(rv reflect.Value) bool {
 	switch rv.Kind() {
 	case reflect.Array, reflect.Slice, reflect.Map, reflect.String:
 		return rv.Len() == 0
 	case reflect.Struct:
-		if rv.Type().Comparable() {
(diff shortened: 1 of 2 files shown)


arp242:
The `isEmptyStruct()` implementation doesn't just check if it's an empty struct, but also if it's an empty array, slice, or map, and the Array implementation is different from `isEmpty`:

	type Struct struct {
		Arr  [3]string `toml:",omitempty"`  // Never empty as len always longer

		// But this behaves different, as it will check the zero value of Arr.
		Nest struct {
			Arr [3]string
		} `toml:",omitempty"`
	}

In principle the second implementation is better, but right now it's documented as this for `Encoder`:

	The toml struct tag can be used to provide the key name; if omitted the
	struct field name will be used. If the "omitempty" option is present the
	following value will be skipped:

	  - arrays, slices, maps, and string with len of 0
	  - struct with all zero values
	  - bool false

I think we can change it without breaking compatibility; now `omitempty` on an array is basically useless, but it should be identical throughout (and the documentation updated, and test added).

I think re-using `isEmpty()` for slices, maps, and arrays rather doing it in `isEmptyStruct()` is better, as it avoids implementations differing.

---

Also, I think the tests should be expanded, for example:

	type Nest struct {
		Field struct {
			uncomp Uncomparable
		}
	}

	type Embed struct {
		Uncomparable
	}

	type EmbedNest struct {
		Field struct {
			Uncomparable
		}
	}

	type UncompSlice struct {
		[]Uncomparable
	}

	type UncompSliceNest struct {
		Field struct {
			uncomp []Uncomparable
		}
	}

arp242:
I think changing the array behaviour would be done best in a different PR btw, and `encoding/json` has the same check:

	case reflect.Array, reflect.Map, reflect.Slice, reflect.String:
		return v.Len() == 0

So I'm not sure yet if we *should* change it.

arp242:
Discussion about that on encoding/json: https://github.com/golang/go/issues/29310

kkHAIKE:
'omitempty' from encoding/json not work on struct type, so isEmpty only check array.len==0 is well.

Actually, original toml's behaviour 'reflect.Zero(rv.Type()).Interface() == rv.Interface()' is same as 'rv.IsZero()', because  'reflect.Zero' is make a zero value struct

I think 'isEmptyStruct' behaviour for Array type struct field is right, because 'reflect.Zero' can make all zero value Array, not len==0 Array.

arp242:
Fixed via #366

PR #370, opened 2022-10-23
Test omitempty for pointers

joukewitteveen:
The value behind the pointer should not matter, just whether the pointer is nil or not.


arp242:
I added a more expansive test for this (with a `t.Skip()` for now) in 68ce1fc.

PR #374, opened 2023-01-10
Fix encoding anonymous struct with multiple fields on 32-bit systems

saracen:
This fixes another 32-bit edge case, which would result in the last field of an anonymous struct being duplicated for every other field within that same struct when encoded.

The fix requires a slice to be copied, which is only needed on 32-bit systems. An existing fix was done for `fieldsDirect`, but this expands to using the copied slice for `fieldsSub` too.

I've added a test case that captures the failure.

On Linux, executing `GOARCH=386 go test -run TestEncodeAnonymousStruct` without the changes done in `encode.go` results in:

```golang
--- FAIL: TestEncodeAnonymousStruct (0.00s)
    --- FAIL: TestEncodeAnonymousStruct/embedded_anonymous_multiple_fields (0.00s)
        encode_test.go:373:
            have:
                [Inner2]
                  N = 4

                [Inner2]
                  N = 4
            want:
                [Inner1]
                  N = 3

                [Inner2]
                  N = 4
```

With the changes, it works as expected.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -457,6 +457,16 @@ func (enc *Encoder) eStruct(key Key, rv reflect.Value, inline bool) {
 
 			frv := eindirect(rv.Field(i))
 
+			if is32Bit {
+				// Copy so it works correct on 32bit archs; not clear why this
+				// is needed. See #314, and https://www.reddit.com/r/golang/comments/pnx8v4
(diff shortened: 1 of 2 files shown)


saracen:
@arp242 Thank you for merging. Do you know when there will be a tagged release with this fix?

PR #381, opened 2023-01-28
skip UTF-8 BOM also

kmuto:
I noticed that reading a UTF-8 encoded file with a BOM causes an error. Unfortunately, major Windows software adds a BOM to UTF-8 files.
UTF-16 support is fixed in #277 , but in addition, skipping the UTF-8 BOM (ef bb bf,  https://en.wikipedia.org/wiki/Byte_order_mark ) also solves the problem, I believe.

A sample is attached.

before:
```
OK(withoutBOM) あ
ERROR(withBOM) toml: line 1: expected '.' or '=', but got '\ufeff' instead
```

after:
```
OK(withoutBOM) あ
OK(withBOM) あ
```

[mini.zip](https://github.com/BurntSushi/toml/files/10527471/mini.zip)

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -66,6 +66,7 @@ func TestDecodeBOM(t *testing.T) {
 	for _, tt := range [][]byte{
 		[]byte("\xff\xfea = \"b\""),
 		[]byte("\xfe\xffa = \"b\""),
+		[]byte("\xef\xbb\xbfa = \"b\""),
 	} {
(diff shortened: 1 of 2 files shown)


arp242:
Thanks!

kmuto:
@arp242 
Thanks for the merge!
Have you decided when you plan to release the next one? If you are keeping the releases at 3 month intervals, I think it's about the right time. 😊

PR #391, opened 2023-05-23
Fix line-ending backslash whitespace escaping for multiline strings

### LINKED ISSUE #372: Line ending backslash in """ causes later line breaks to also be removed
Hi, I'm having trouble with breaking long lines in `"""` delimited strings using a backslash. This is called a "line ending backslash" in the toml spec.

Reproducer:

```go
package main

import (
	"fmt"
	"os"

	"github.com/BurntSushi/toml"
)

const in = `f = """
a\
b
c
"""`

func main() {
	var out map[string]string
	if _, err := toml.Decode(in, &out); err != nil {
		fmt.Fprintf(os.Stderr, "error: %v\n", err)
		os.Exit(1)
	}

	fmt.Println(out["f"])
}
```

Expected output, which is generated with BurntSushi/toml v0.3.1:
```
ab
c

```

Output with BurntSushi/toml v1.2.1:
```
abc
```

Any chance of fixing this? Thanks!

cespare:
Fixes #372.

### DIFF
--- a/internal/toml-test/tests/valid/string/multiline.json
+++ b/internal/toml-test/tests/valid/string/multiline.json
@@ -50,5 +50,9 @@
   "whitespace-after-bs": {
     "type": "string",
     "value": "The quick brown fox jumps over the lazy dog."
+  },
+  "only-ignore-first": {
+    "type": "string",
+    "value": "Here are two\nlines of text.\nAnd another\n  two.\n"
   }
 }
--- a/internal/toml-test/tests/valid/string/multiline.toml
+++ b/internal/toml-test/tests/valid/string/multiline.toml
@@ -35,6 +35,16 @@ whitespace-after-bs = """\
        the lazy dog.\	
        """
 
+only-ignore-first = """
+Here are \
+  two
+lines of text.
+And \
+
+  another
+  two.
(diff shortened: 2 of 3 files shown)


arp242:
It would be better to put the test for this in [toml-test](https://github.com/BurntSushi/toml-test), rather than adding a new `TestDecodeMultilineStringBackslash` for this – that way other TOML implementations benefit from this too, and it's also easier.

There's a copy of that in `internal` here, and it gets tested with just `go test` (no need to run the `toml-test` binary like before) – can just modify that and I'll sync toml-test with it at some point.

Specifically: https://github.com/BurntSushi/toml/blob/master/internal/toml-test/tests/valid/string/multiline.toml (and the corresponding .json file).

cespare:
> can just modify that and I'll sync toml-test with it at some point.

Interesting -- this repo is the source of truth for toml-test? I had assumed it was the other way around.

arp242:
> Interesting -- this repo is the source of truth for toml-test? I had assumed it was the other way around.

It is; I just didn't want to bother you with having to create two PRs.

cespare:
OK, thanks! Done.

PR #405, opened 2024-05-02
Add Marshal Function

samawise:
Added the Marshal function which returns the TOML representation of the Go value as bytes along with any error that may occur while marshaling. This adds parity with the existing Unmarshal function and conforms to the standard libraries JSON package, which also has both Unmarshal and Marshal.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -2,6 +2,7 @@ package toml
 
 import (
 	"bufio"
+	"bytes"
 	"encoding"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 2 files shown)


arp242:
Seems okay to add; can you add a basic test? Or maybe just modify one of the calls in `TestEncodeRoundTrip()`. Just to ensure the function is called and works – doesn't need to be comprehensive or anything.

samawise:
Wrote a test for it under `TestEncodeRoundTrip()`.

arp242:
Cheers, thanks

PR #406, opened 2024-05-03
fuzz: move fuzz_targets from oss-fuzz

manunio:
- Moves fuzz_targets from oss-fuzz. This makes it easy to maintain the fuzzers and minimizes breakages that can arise as source code changes over time.

- Adds cifuzz action workflow which is a service provided by oss-fuzz where this project already runs, this helps in catching shallow bugs, regression or build breakage by running fuzzers on PR for ~5 minutes.


arp242:
Thanks!

PR #419, opened 2024-07-09
Fix metadata - alternate approach

### LINKED ISSUE #417: metadata.Keys (and thus metadata.Undecoded) is invalid in some cases
**Sample code:**
```
package main

import (
	"fmt"

	"github.com/BurntSushi/toml"
)

func main() {
	content := `[table.subtable.subsubtable]
a = 1
b = 2
`
	var v interface{}
	metadata, err := toml.Decode(content, &v)
	if err != nil {
		panic(err)
	}
	fmt.Printf("%#v\n", metadata.Keys())
}
```

Playground: https://go.dev/play/p/Rr5bm5DbEql

**Output:**
```
[]toml.Key{toml.Key{"table", "subtable", "subsubtable"}, toml.Key{"table", "subtable", "subsubtable", "b"}, toml.Key{"table", "subtable", "subsubtable", "b"}}
```

Notice how the second and third value in that slice both point to `b`.

**Expected output:**
```
[]toml.Key{toml.Key{"table", "subtable", "subsubtable"}, toml.Key{"table", "subtable", "subsubtable", "a"}, toml.Key{"table", "subtable", "subsubtable", "b"}}
```

The root cause seems to be somewhere near this [append call](https://github.com/BurntSushi/toml/blob/eb727477b3f7e4aa878635d39257753f0840811b/meta.go#L139). The result of `append` can return a slice backed by the same array if the capacity is available. See Go example code: https://go.dev/play/p/2FGKuCvCUKz.

maroux:
Alternative approach to #418.

[This line](https://github.com/BurntSushi/toml/blame/eb727477b3f7e4aa878635d39257753f0840811b/meta.go#L139) added in #400 caused the problem since the result of multiple `append` calls can return two slices backed by the same backing array. See demo here: https://go.dev/play/p/2FGKuCvCUKz. The bug only surfaces under specific conditions when the capacity of `parser.context` is increased by more than 1 due to [this append call](https://github.com/BurntSushi/toml/blob/eb727477b3f7e4aa878635d39257753f0840811b/parse.go#L606). This PR removes the `append` calls that cause this issue, as well as optimizing some places where you don't need to `append` since the length is predetermined -- that is, instead of:
```
a = make([]string, 0, N)
for i := 0; i < N; i++ {
        a = append(a, value)
}
```
prefer:
```
a = make([]string, N)
for i := 0; i < N; i++ {
        a[i] = value
}```

Fixes #417


arp242:
> For the record, I think #418 is the better fix 

Yes I agree. Closing this. I don't think I added that for a hugely specific reason other than reading the code while fixing that bug and thinking "we can optimize this a bit".

PR #421, opened 2024-08-01
Return `[]any` for hash arrays when decoding to map / any.

veigaribo:
Currently they are returned in `[]map[string]any`, which makes the tree impractical to traverse.

Example: <https://github.com/itchyny/gojq/blob/0607aa5af33a4f980e3e769a1820db80e3cc7b23/encoder.go#L72> triggers the panic, instead of being encoded as an array.

Before my changes, the library depended on the peculiar behaviour of hash arrays and unfinished inline arrays being `[]map[string]any` while finished inline arrays became `[]any`. That was relevant for preventing finished inline arrays from being added to.

Since now everything is `[]any`, I added `locked` to the key info once the inline array (or any top-level field, really) is finished in order to be able to tell.

I'm not happy about the duplicate error lines, but am not sure it would be better to engineer something just for that.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -642,9 +642,9 @@ func (d *dish) UnmarshalTOML(p any) error {
 	data, _ := p.(map[string]any)
 	d.Name, _ = data["name"].(string)
 	d.Price, _ = data["price"].(float32)
-	ingredients, _ := data["ingredients"].([]map[string]any)
+	ingredients, _ := data["ingredients"].([]any)
 	for _, e := range ingredients {
-		n, _ := any(e).(map[string]any)
+		n, _ := e.(map[string]any)
 		name, _ := n["name"].(string)
 		i := ingredient{name}
 		d.Ingredients = append(d.Ingredients, i)
--- a/internal/tag/add.go
+++ b/internal/tag/add.go
@@ -27,12 +27,6 @@ func Add(key string, tomlData any) any {
 
 	// An array: we don't need to add any tags, just recurse for every table
 	// entry.
-	case []map[string]any:
-		typed := make([]map[string]any, len(orig))
-		for i, v := range orig {
-			typed[i] = Add("", v).(map[string]any)
-		}
-		return typed
 	case []any:
 		typed := make([]any, len(orig))
 		for i, v := range orig {
--- a/internal/toml-test/toml.go
+++ b/internal/toml-test/toml.go
@@ -38,12 +38,6 @@ func (r Test) CompareTOML(want, have any) Test {
 	switch w := want.(type) {
 	case map[string]any:
 		return r.cmpTOMLMap(w, have)
-	case []map[string]any:
-		ww := make([]any, 0, len(w))
-		for _, v := range w {
-			ww = append(ww, v)
-		}
-		return r.cmpTOMLArrays(ww, have)
 	case []any:
 		return r.cmpTOMLArrays(w, have)
 	default:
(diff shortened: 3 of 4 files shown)


arp242:
In principle this looks good at a glance. My main concern is that it might break compatibility where people use the type in some way; that's also why I haven't changed it before. I'm not entirely sure if that fear is warranted though.

veigaribo:
Agree. It would be safest to put this in a new major version as a breaking change, if you guys are willing to do that. Even if I don't think many people will be affected by it, judging by the lack of related reported issues (that I could find at least).

About the fuzzer, I'm a bit confused. Is it failing to parse / decode `l=[{''.''=1},"",{''.''.''.''.''.''.''.''.''.''.''.''=0}]`?

In my machine it seems to work fine:

```go
package main

import (
	"fmt"
	"encoding/json"

	"github.com/BurntSushi/toml"
)

func main() {
	t := `l=[{''.''=1},"",{''.''.''.''.''.''.''.''.''.''.''.''=0}]`

	var data any
	toml.Unmarshal([]byte(t), &data)

	pp, _ := json.Marshal(data)
	fmt.Println(string(pp))
}
```

Outputs

```
veigaribo
{"l":[{"":{"":1}},"",{"":{"":{"":{"":{"":{"":{"":{"":{"":{"":{"":{"":0}}}}}}}}}}}}]}
```

I have a `go.mod` with `replace github.com/BurntSushi/toml v1.4.0 => ../toml`, pointing to my code, in my branch, and I added the `fmt.Println("veigaribo")` to make sure it's working.

arp242:
It's this bit that fails: https://github.com/BurntSushi/toml/blob/master/ossfuzz/fuzz.go#L36

It's the decode + encode + decode that errors, not just a decode. Here's a test case:

```
diff --git i/decode_test.go w/decode_test.go
index 5b43730..cdf97ca 100644
--- i/decode_test.go
+++ w/decode_test.go
@@ -1144,3 +1144,20 @@ func BenchmarkKey(b *testing.B) {
                k.String()
        }
 }
+
+func TestXXX(t *testing.T) {
+       doc := `l=[{''.''=1},"",{''.''.''.''.''.''.''.''.''.''.''.''=0}]`
+       var m any
+       _, err := Decode(doc, &m)
+       fmt.Println(err)
+       fmt.Println(m)
+
+       out, err := Marshal(m)
+       fmt.Println(err)
+       fmt.Println(string(out))
+
+       var m2 any
+       _, err = Decode(string(out), &m2)
+       fmt.Println(err)
+       fmt.Println(m2)
+}
```

On your branch:

```
[~cg/toml](slice-of-any)% go test -run TestXXX
<nil>
map[l:[map[:map[:1]]  map[:map[:map[:map[:map[:map[:map[:map[:map[:map[:map[:map[:0]]]]]]]]]]]]]]
<nil>
l = [{"" = {"" = 1}}, "", {"" = {"" = {"" = {"" = {"" = {"" = {"" = {"" = {"" = {"" = {"" = {"" = 0}}}}}}}}}}}}]

toml: line 1: Key 'l' was already created and cannot be used as an array.
<nil>
PASS
ok      github.com/BurntSushi/toml      0.002s
```

On master:

```
[~cg/toml](master)% go test -run TestXXX
<nil>
map[l:[map[:map[:1]]  map[:map[:map[:map[:map[:map[:map[:map[:map[:map[:map[:map[:0]]]]]]]]]]]]]]
<nil>
l = [{"" = {"" = 1}}, "", {"" = {"" = {"" = {"" = {"" = {"" = {"" = {"" = {"" = {"" = {"" = {"" = 0}}}}}}}}}}}}]

<nil>
map[l:[map[:map[:1]] map[:map[:map[:map[:map[:map[:map[:map[:map[:map[:map[:map[:0]]]]]]]]]]]]]]
PASS
ok      github.com/BurntSushi/toml      0.002s
```

I haven't looked into it why this fails beyond reproducing it. Regardless of what we do with respect of compatibility, that should probably work.

veigaribo:
Thank you. One more TODO gone.

Had to create an specific function for setting the types of hashes because `setType` was being invoked with `""` for empty keys and, due to the previous conditional, caused the type of the parent array to be set and locked instead of itself. I believe. Weird how that worked before.

PR #422, opened 2024-08-01
Allow for setting TOML next.

veigaribo:
Currently, the only way to enable TOML next (1.1.0) features is via the environment variable, which is very ugly if you want to do it in application code. Therefore I add the `SetTomlNext` method to the Decoder so it can be done more cleanly.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -118,7 +118,8 @@ const (
 // This decoder does not handle cyclic types. Decode will not terminate if a
 // cyclic type is passed.
 type Decoder struct {
-	r io.Reader
+	r        io.Reader
+	tomlNext *bool
 }
 
 // NewDecoder creates a new Decoder.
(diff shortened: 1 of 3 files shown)


arp242:
I would not be in favour of this because we would never be able to remove that method as that would cause code to not compile. That's why it's an environment variable. That's also why 

I agree that using os.Setenv() in your application code is a bit ugly, but I think it's not that big of a deal – it's just one line.

veigaribo:
I don't think we would ever need to remove it, though. It has the semantics of enabling whatever features have been approved by the TOML team but have not yet been put in a proper release (and we have implemented, of course). Even in the scenario where development of TOML has ceased completely, there will be no such features, so this shouldn't, and wouldn't, do anything. And that sounds fine to me.

However, now that I think about it, build tags seem more appropriate for this...

arp242:
> However, now that I think about it, build tags seem more appropriate for this...

That will require rebuilding the application. A second advantage of environment variables I forgot to mention before is that users can set it without recompiling.

Ideally the TOML spec shouldn't be as dead as it is and we shouldn't need any method to enable unreleased features, but yeah ... I can't really change that... :-( And after TOML 1.1 there probably won't be any further changes, assuming we will ever see a 1.1 release in the first place.

I really appreciate the time you took to write this patch, but I think os.Setenv() is fine and I really don't want to be lumbered with an extra method that will be a no-op, so I'm afraid I'll have to reject it unless there's a good object argument in favour of something else ("objective" in the sense of "I want to do X, but it's impossible now").

Issue #425, opened 2024-09-24
Custom UnmarshalTOML() decoded keys left in meta.Undecoded() 


mvo5:
First, thanks your amazing library!

It looks like when doing a custom UnmarshalTOML() for a struct this will leave the keys decoded there in meta.Undecoded().

Here is a minimal test case:
```go
type CustomStruct struct {
	Foo string `json:"foo"`
}

func (cs *CustomStruct) UnmarshalTOML(data interface{}) error {
	d, _ := data.(map[string]interface{})
	cs.Foo = d["foo"].(string)
	return nil
}

func TestDecodeCustomStruct(t *testing.T) {
	var cs CustomStruct
	meta, err := Decode(`
		foo = "bar"
	`, &cs)
	if err != nil {
		t.Fatalf("Decode failed: %s", err)
	}

	if cs.Foo != "bar" {
		t.Errorf("\nhave:\n%v\nwant:\n%v\n", cs.Foo, "bar")
	}
	if len(meta.Undecoded()) > 0 {
		t.Errorf("\ncustom decode leaves unencoded fields: %v\n", meta.Undecoded())
	}
}
```
and the output:
```
$ go test -run TestDecodeCustomStruct
--- FAIL: TestDecodeCustomStruct (0.00s)
    decode_test.go:1174: 
        custom decode leaves unencoded fields: [foo]
``` 
Our use-case is very similar [0], we want to dynamically handle int or string sizes in the toml.

I think we can fix this particular issue via moving decoding to primitive types and not structs
but it would be nice to have a way in  UnmarshalTOML() for complex types to signal to the
MetaData object that certain keys are decoded (or maybe this exists and I just couldn't
figure it out)?

[0] https://github.com/osbuild/images/blob/main/pkg/blueprint/filesystem_customizations.go#L27

arp242:
This seems like a reasonable change.

Could you add a more extensive test though? At the very least something like:

	a   = 1
	arr = [2]

	[tbl]
	b = 3

	inline = {c = 4}

Just so that keys in different locations are tested: array, table, inline table. Maybe a few more? I'm not sure if these are different code-paths off-hand, but doesn't hurt to test it.

PR #426, opened 2024-09-24
Set mapping keys to "decoded" when custom unmarshaler is used

mvo5:
decode: set all keys under a mapping with custom marshaler
 decoded

This is a naive fix for the issue of custom unmarshal and marking
keys as "decoded". It simply assumes that if there was a custom
unmarshal for a mapping type all keys got handlded by the custom
unmarshal code.

This might be too naive but it seems reasonable and it also seems
we would need a richer interface for a custom unmarshal that gives
access to `MetaData` if we want to be more fine grained.

---

decode: add (failing) test for undecoded fields in a struct

This commit adds a failing test for the scenario decribed in issue
https://github.com/BurntSushi/toml/issues/425

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -196,6 +196,19 @@ func (md *MetaData) PrimitiveDecode(primValue Primitive, v any) error {
 	return md.unify(primValue.undecoded, rvalue(v))
 }
 
+// markDecodedRecursive is a helper to mark any key under the given tmap as
+// decoded, recursing as needed
+func markDecodedRecursive(md *MetaData, tmap map[string]any) {
+	for key := range tmap {
+		md.decoded[md.context.add(key).String()] = struct{}{}
+		if tmap, ok := tmap[key].(map[string]any); ok {
+			md.context = append(md.context, key)
+			markDecodedRecursive(md, tmap)
+			md.context = md.context[0 : len(md.context)-1]
+		}
+	}
+}
+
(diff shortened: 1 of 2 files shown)


arp242:
This seems like a reasonable change.

Could you add a more extensive test though? At the very least something like:

	a   = 1
	arr = [2]

	[tbl]
	b = 3

	inline = {c = 4}

Just so that keys in different locations are tested: array, table, inline table. Maybe a few more? I'm not sure if these are different code-paths off-hand, but doesn't hurt to test it.

mvo5:
Thanks a lot for your feedback! I updated the code and it now includes the more complex testcase that you suggested and that uncovered missing handling of nested which I added now too. I am happy to add more tests, but I need to think a bit what else to check for (I'm not super familiar with the toml details). Ideas welcome here of course :) 

Please also let me know if I should squash or rebase my commits, I did them linearly mostly to make it easy to review but happy to do whatever is needed/preferred.

mvo5:
Friendly ping, anything I can do or improve to help this along?

arp242:
I'd like to carefully look at it and play around with it a bit. Not terrible time-consuming, but I don't have a lot of bandwidth for that right now. It's "on the list" and not forgotten, but it might be a bit before I have a chance to properly look at it.

mvo5:
Thank you so much! :heart:

PR #427, opened 2024-10-08
godocs.io isn't working

SvenDowideit:
I don't know if this is just a temporary thing, but the existing docs link isn't working today :)


arp242:
Well, I guess it's down for good; a shame, because I much prefer it over pkg.go.dev :-(

PR #433, opened 2024-11-22
Fix ErrorWithPosition panic when less than two lines

xoltia:
Currently, it is possible for `ParserError.ErrorWithPosition` to cause a panic for some input.
You can reproduce the panic using this example:

```go
v := &struct{ String string }{}
_, err := toml.Decode(`string = "test`, v)
parseErr := err.(toml.ParseError)
_ = parseErr.ErrorWithUsage()
``` 

I've added a fix and a new test case which covers this error.

### DIFF
--- a/error_test.go
+++ b/error_test.go
@@ -245,6 +245,19 @@ func TestParseError(t *testing.T) {
             |         15:04:05.856018510
 			`,
 		},
+
+		{
+			&struct{ String string }{},
+			`string = "test`,
+			`
+            | toml: error: unexpected EOF; expected '"'
+			|
+			| At line 0, column 14:
+			|
+			|       0 | string = "test
+			|                        ^
+			`,
+		},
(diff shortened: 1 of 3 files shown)


error_test.go
```diff
@@ -245,6 +245,19 @@ func TestParseError(t *testing.T) {
...
+
+		{
+			&struct{ String string }{},
+			`string = "test`,
+			`
+            | toml: error: unexpected EOF; expected '"'
+			|
+			| At line 0, column 14:
+			|
+			|       0 | string = "test
```
arp242, line 257: This is line 1, not 0.

xoltia: Sorry, I didn't notice that. I've updated the test and believe I found where the line number was being incorrectly set to 0.

arp242:
Merged via 4664468; thanks!

PR #440, opened 2025-04-01
decode: mark nested fields as decoded as well

mvo5:
This commit is a small followup for PR#426 to handle another case with the custom unmarshal. This was found in a complex TOML based image customization "blueprint", c.f. https://github.com/osbuild/blueprint/pull/12

With this commit the test in the other PR passes but I also extracted a minimal testcase (the naming may need tweaking, I'm not super familar with the naming convention inside "toml")


arp242:
Cheers, thanks

mvo5:
Thank you so much, really appreciated!

Issue #443, opened 2025-05-30
No way of skipping a field when marshaling


lzap:
Hello, I need my custom `MarshalTOML` method to be able to omit a field completely when it is set to `nil` but it does not appear to be possible. It must not return. `nil, nil` that is an error, there is no special error value for omitting a field, returning an empty buffer creates an invalid TOML:

```toml
# this is not valid
field_that_needs_to_be_omitted = 
```

My type implements the Go 1.24 newly added `IsZero()` method but I guess that is something that is not supported in this library. That could be perhaps a solution.

Is there a solution? I do not have access to the script that encapsulates the type. Thanks.

PR #444, opened 2025-05-30
Tag omitzero respects IsZero()

lzap:
This patch adds support for `IsZero` method semantics [introduced in Go 1.24](https://tip.golang.org/doc/go1.24) for JSON. When a type has this method implemented as `func (T) IsZero() bool` and it returns true and TOML Go struct `omitzero` exists, the library will omit it completely.

This introduces a change, Go built-in type `time.Time` already has this method and previously when an empty time was marshaled it created TOML `time = 0001-01-01T00:00:00Z` while now this value is skipped. But this will be only done for those fields which are tagged with `omitzero` which is not an issue, no?

Solves: https://github.com/BurntSushi/toml/issues/443

Note this feature does not need Go 1.24, it only follows what Go team is doing for JSON but this will work with any Go version supported by `toml`.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -663,7 +663,34 @@ func getOptions(tag reflect.StructTag) tagOptions {
 	return opts
 }
 
+func callIsZero(rv reflect.Value) (bool, bool) {
+	if !rv.IsValid() {
+		return false, false
+	}
+
+	isZeroMethod := rv.MethodByName("IsZero")
+	if !isZeroMethod.IsValid() {
+		return false, false
+	}
+
+	methodType := isZeroMethod.Type()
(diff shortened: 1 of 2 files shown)


lzap:
Oh so is my understanding that my patch is completely useless and the library already does this? Wow, this is not the first time this happened to me. I can close then.

arp242:
Not entirely; ideally I would like this to behave identical to stdlib. I'm just not entirely sure what the best path for that is 🤔 I'd like to avoid a v2 if at all possible.

Just saying that you shouldn't *need* this patch for your particular issue.

lzap:
Okay let me know if you want me to rewrite the patch.

PR #447, opened 2025-07-03
Adds support for IsZero method

MattWhelan:
Also adds some surprising tests related to type aliases implementing interfaces.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -65,6 +65,10 @@ var dblQuotedReplacer = strings.NewReplacer(
 	"\x7f", `\u007f`,
 )
 
+type zeroer interface {
+	IsZero() bool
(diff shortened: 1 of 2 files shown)


MattWhelan:
Related to #443. 

The runtime type behavior illustrated by the tests here is truly odd, but may help explain the trouble people have had with Marshaller implementations and type aliases.

arp242:
Isn't this the same as #444?

MattWhelan:
> Isn't this the same as #444?

Similar, certainly. The tests are worth a look, even if for some reason we can't merge the actual change. The odd interaction of `CanAddr` and custom marshalling for type aliases is really surprising.

PR #455, opened 2025-12-15
reject duplicate array

### LINKED ISSUE #454: duplicate arrays should be rejected
The decoder should probably reject TOML like this:
```toml
bad = [1,2]
bad = [3,4]
```
With v1.5.0, there is no error and the first is discarded, the final value for `bad` is `[3,4]`.

alexander-buerger-met-no:
Closes #454.


alexander-buerger-met-no:
Thanks, I was not aware of the tom-lang/toml-test repo.

Issue #449, opened 2025-07-24
case-duplicate key stores random value in struct


alexander-buerger-met-no:
# Summary
When reading TOML files with keys that differ in case into a struct, the value stored seems to be one of the values given, apparently randomly. As the case is different, this is valid TOML. But `github.com/BurntSushi/toml` seems to treat them as identical keys without complaining about duplicates. I did not figure out how to avoid that. Explicitly setting `toml:"fish"` as struct tag did not help.

# Example
```
package main

import (
	"fmt"
	"log"

	"github.com/BurntSushi/toml"
)

func main() {
	var blob = `
		fish = "Cod"
		Fish = "Shark"
	`

	var conf struct {
		Fish string `toml:"fish"`
	}
	_, err := toml.Decode(blob, &conf)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Printf("%s\n", conf.Fish)
}
```
## Expected
Some kind of error, or at least consistently printing Cod or Shark.

## Observed
Randomly prints Cod or Shark.

alexander-buerger-met-no:
I made a PR with a possible solution, just a very small change (1 line) plus a test case (22 lines).

PR #458, opened 2025-12-19
add test for case-sensitive toml keys from struct tags

alexander-buerger-met-no:
Makes toml keys case sensitive iff the key is specified in a struct tag. This might be a solution for #449.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -313,7 +313,7 @@ func (md *MetaData) unifyStruct(mapping any, rv reflect.Value) error {
 				f = ff
 				break
 			}
-			if f == nil && strings.EqualFold(ff.name, key) {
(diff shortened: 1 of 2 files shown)


arp242:
This is not a compatible change; something like this:

    var m struct {
        Fish string `toml:"fish"`
    }
    _, err := Decode(`FISH = "cod"`, &m)
    fmt.Println(err, m)

Will currently (consistently) output `<nil> {cod}`, but with this change it outputs `<nil> {}`.

You can argue the new behaviour is better, but it's been like this for about 15 years, people do make typos, and I don't want to break existing programs.

The only fix we can do for v1 is to consistently pick one or the other if you have two keys that differ only in casing, but we can't make it case-sensitive. Like I said in the issue: I don't think it's really worth the effort given that it's been like this for a very long time and I'll change it in v2 once Go 1.27 is released with encoding/json/v2.

PR #461, opened 2025-12-25
New EncodeFile method with a test

ReviStudios:
## What is added: 
This adds a `EncodeFile` method similar to the `DecodeFile` method. Instead of reading a file, it writes to one and returns an error if anything goes wrong. 

**Use case:**
This function should make it easier to for example save your config data to your `config.toml` file without writing your own function.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"os"
 	"reflect"
 	"sort"
 	"strconv"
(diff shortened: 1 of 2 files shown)


arp242:
I think I would rather move towards a simpler API, similar to encoding/json and encoding/json/v2 instead of adding functions.

Adding O_TRUNC can be pretty surprising and destroy data. But then again, removing it and now having it work can also be surprising. If encoding the file fails you're left with an empty file. Ideally overwriting existing files should be atomic, where it writes to a temporary file and then moves it.

Thanks for your effort, but I think this sort of thing is probably better left up to the caller.

ReviStudios:
You're right, thanks for mentioning the reason :)

PR #471, opened 2026-04-29
fix: parsing TOML tags with commas in key names, add regression tests

drupol:
Hello,

This PR fix struct tag parsing so TOML key names containing commas are preserved during encode and decode.

This addresses cases like `stack(x,n)`, which were previously truncated at the comma (see https://github.com/sclevine/yj/issues/52)

The tag parser split on every comma and treated the first segment as the key name.

Example behavior before this change:
- Tag value: `stack(x,n),omitempty`
- Parsed key name: `stack(x`
- Result: incorrect key lookup/output

Before this PR, tag parsing used a global comma split, which cannot distinguish:
- commas that are part of the key name
- commas that introduce options such as `omitempty` and/or `omitzero`

With this PR, tag parsing has been updated:
- parse known options from the right side only
- keep the remaining left side as the full key name
- preserve unknown suffixes as part of the key name instead of truncating

This keeps existing option behavior while correctly handling commas in key names.

I also added regression tests for both encode and decode.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -1064,6 +1064,20 @@ func TestDecodeDoubleTags(t *testing.T) {
 	}
 }
 
+func TestDecodeTagNameWithComma(t *testing.T) {
+	var s struct {
+		Value string `toml:"stack(x,n)"`
+	}
+	_, err := Decode(`"stack(x,n)" = "foobar"`, &s)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if s.Value != "foobar" {
+		t.Errorf("\nhave: %s\nwant: %s\n", s.Value, "foobar")
+	}
+}
+
 func TestMetaKeys(t *testing.T) {
 	tests := []struct {
 		in   string
--- a/encode.go
+++ b/encode.go
@@ -650,16 +650,29 @@ func getOptions(tag reflect.StructTag) tagOptions {
 		return tagOptions{skip: true}
 	}
 	var opts tagOptions
-	parts := strings.Split(t, ",")
-	opts.name = parts[0]
-	for _, s := range parts[1:] {
+	for {
+		i := strings.LastIndexByte(t, ',')
+		if i < 0 {
+			break
+		}
+
+		s := t[i+1:]
 		switch s {
 		case "omitempty":
 			opts.omitempty = true
+			t = t[:i]
 		case "omitzero":
 			opts.omitzero = true
+			t = t[:i]
+		default:
+			// Unknown suffix: treat the full tag value as the key name.
+			i = -1
+		}
+		if i < 0 {
+			break
 		}
 	}
+	opts.name = t
 	return opts
 }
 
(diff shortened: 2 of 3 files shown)


arp242:
I don't want to merge this as I want to keep this as similar to stdlib as possible; encoding/json and encoding/json/v2 also doesn't allow this. The syntax for struct tags is `name[,options]` with no way to escape the `,`. Inventing our own format doesn't strike me as a good idea, especially not doing it this was where it's unclear if `x,y` is a name or name,option pair.

So this is something that needs to be fixed in the Go canonical struct tag format (e.g. https://github.com/golang/go/issues/15000). There has also been some movement for typed struct tags recently (https://github.com/golang/go/issues/74472) that would solve this.

drupol:
Hey,

Thanks for the feedback, that makes sense regarding keeping behavior close to `stdlib`, I fully agree.

The concrete issue I am trying to fix here is the silent truncation, **not to introduce escaping or a brand new tag format**.

Today, with a tag like `toml:"stack(x,n)"`, the parser effectively interprets it as `name = "stack(x"` and an extra suffix `"n)"` (_basically, any tag containing a comma_). That means encode/decode can target the wrong key without any explicit error.

So the bug is:
1. input expresses one key
2. parser silently rewrites it to another key
3. result is incorrect lookup/output with no signal

Once again, I agree this should stay aligned with Go tag conventions. If supporting commas in names is out of scope, I can rework the PR to fail fast instead (explicitly reject ambiguous tags) rather than silently truncating them. I think that would still fix the real bug: silent wrong behavior.

arp242:
All the more reason to not change it as it would be an incompatible change.

And it's also how JSON works:

	package main

	import (
		"encoding/json"
		json2 "encoding/json/v2"
		"fmt"
	)

	func main() {
		var x struct {
			Field string `json:"stack(x,y)"`
		}
		err := json.Unmarshal([]byte(`{"stack(x,y)": "abc"}`), &x)
		fmt.Println(err) // <nil>
		fmt.Println(x)   // {}

		err = json.Unmarshal([]byte(`{"stack(x": "def"}`), &x)
		fmt.Println(err) // <nil>
		fmt.Println(x)   // {def}

		err = json2.Unmarshal([]byte(`{"stack(x,y)": "ghi"}`), &x)
		fmt.Println(err) // json: Go struct field Field has malformed `json` tag: invalid character ')' before next option (expecting ',')
		fmt.Println(x)   // {def}
	}
		fmt.Println(err)
		fmt.Println(x)
	}

Which outputs:

	<nil>
	{}
	<nil>
	{def}
	json: cannot unmarshal JSON object into Go struct { Field string "json:\"stack(x,y)\"" }: Go struct field Field has malformed `json` tag: invalid character ')' before next option (expecting ',')
	{def}

Note that json/v2 only errors because of the `)`; with e.g. `json:"stack,xy"` it also truncated.

Again, this is something you need to take up on the Go repo. Last week Austin said the typed struct tag issue has some priority due to encoding/v2.

drupol:
Fair enough, thanks for the detailed explanation.

PR #472, opened 2026-05-04
errors: implement Unwrap for ParseError

cyphar:
For types with custom UnmarshalTOML methods, being able to return an
error that is then checked using errors.Is() is quite handy and matches
user expecations with regards to how error wrapping works in modern Go.

Signed-off-by: Aleksa Sarai <aleksa@amutable.com>

### DIFF
--- a/error.go
+++ b/error.go
@@ -98,6 +98,10 @@ func (pe ParseError) Error() string {
 	return fmt.Sprintf("toml: line %d (last key %q): %s", pe.Position.Line, pe.LastKey, pe.Message)
 }
 
+func (pe ParseError) Unwrap() error {
+	return pe.err
+}
+
 // ErrorWithPosition returns the error with detailed location context.
 //
 // See the documentation on [ParseError].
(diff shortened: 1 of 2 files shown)


cyphar:
I could not find any existing discussion about implementing `Unwrap` in PRs or issues, so I just went ahead and implemented it. I am not aware of any obvious compatibility issue this could introduce, but I guess it is theoretically possible.

arp242:
What concrete problem does this solve? I think this only ever set for internal (unexported) errors?

cyphar:
If you write a custom `UnmarshalTOML`, you can return any error you like and it does get propagated (as you can see in the updated test).

In my particular case, I only needed to check the unwrapped error for testing purposes (to see which error case was triggered) and so I just worked around it using string-based matching, but you could imagine having an `UnmarshalTOML` that has a sentinel error that would trigger re-parsing (or even changing how the error gets displayed), and being able to use `errors.Is` is more ergonomic than string matching.

I'm not particularly fussed about this, but being able to unwrap your own errors is something I would expect to be possible with modern Go projects.

PR #474, opened 2026-05-11
Add UnifyHook for Decoder

crwnet:
This allows users to intercept and customize value transformation before the unify step during decoding.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -21,11 +21,17 @@ type Unmarshaler interface {
 	UnmarshalTOML(any) error
 }
 
+type UnifyHook func(data any, rv reflect.Value) error
+
 // Unmarshal decodes the contents of data in TOML format into a pointer v.
 //
 // See [Decoder] for a description of the decoding process.
 func Unmarshal(data []byte, v any) error {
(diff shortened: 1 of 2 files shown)


arp242:
What does this accomplish that e.g. toml.Unmarshaler or encoding.TextUnmarshaler doesn't?

crwnet:
I use it together with other plugins to provide a unified extension mechanism for nested structs and array types, such as tag-based default value injection and generic initialization of interface. This helps avoid implementing UnmarshalTOML or duplicating parsing logic for every concrete type.

In essence, it acts as a pluggable “generic initialization / enhancement layer” on top of the deserialization process, rather than replacing type-specific mechanisms like toml.Unmarshaler or encoding.TextUnmarshaler.

arp242:
That's not very concrete; I don't follow what exactly this makes easier?

Either way, this doesn't seem to add something that's currently impossible and having 2 ways to do the same is generally a bad thing. In general the plan is to make this compatible with encoding/json/v2 once it (finally) gets finalised, and that doesn't have any of this.

Issue #470, opened 2026-04-22
Adding keys to inline arrays and tables is incorrectly permitted


jmillikin:
These two TOML documents are accepted, but per the spec ought to be rejected:

```toml
a = []
a.b = ''
```

```toml
a = {}
a.b = ''
```

edit: seems to be a more general problem with tracking whether a table was added implicitly or not. This document is also incorrectly accepted:

```toml
[a.b]
[a]
b.c = ''
```

PR #482, opened 2026-06-02
Fix issue #470: Reject adding keys to inline arrays and tables

vigneshakaviki:
## Summary
- Fixed validation bug: parser accepted invalid TOML documents
- Inline arrays and tables cannot have keys added per TOML spec
- Added inline tracking to keyInfo struct to distinguish inline vs header values
- Added validation in topLevel() to reject key extensions on inline values

## Test Plan
- [x] Two new test cases verify rejection of invalid inline array/table extensions
- [x] All existing tests pass
- [x] Error messages are clear and include offending keys

### DIFF
--- a/issue_470_test.go
+++ b/issue_470_test.go
@@ -1,0 +1,65 @@
+package toml
+
+import (
+	"testing"
+)
+
+// Test case 1: Adding keys to inline array
+func TestIssue470InlineArrayInvalidAddKeys(t *testing.T) {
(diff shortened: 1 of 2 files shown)


arp242:
This is not only a bad way to fix it, it is *very obviously* a bad way to fix it if you had spent more than 3 seconds looking at it. Please don't spam this kind AI nonsense.

PR #483, opened 2026-06-02
reject negative integers when decoding into uint and uint64

sahvx655-wq:
unifyInt only checks num < 0 for the uint8/uint16/uint32 cases, so a negative TOML integer decoded into a uint or uint64 field wraps silently instead of erroring: A = -1 into uint64 gives 18446744073709551615. Move the num < 0 check so it covers every unsigned width.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -512,9 +512,10 @@ func (md *MetaData) unifyInt(data any, rv reflect.Value) error {
 		rv.SetInt(num)
 	case rvk >= reflect.Uint && rvk <= reflect.Uint64:
 		unum := uint64(num)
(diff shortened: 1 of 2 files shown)


arp242:
Do you have a test case where this fails?

sahvx655-wq:
Yeah. `A = -1` into a `uint64` field returns no error and leaves `A = 18446744073709551615` on master; same for `uint`. Added both to TestDecodeErrors in 272d5ca, and confirmed they fail (err is nil) without the decode.go change.

arp242:
Cheers, thanks!

Issue #451, opened 2025-08-24
Re-definition of direct ancestor may pass undetected


lyokha:
Consider a test:

```go
package main

import (
	"log"
	"github.com/BurntSushi/toml"
) 

type Test struct {
	a A
}

type A struct {
	b string
	c string
}

func main() {
	var test Test
	_, err := toml.DecodeFile("./test.toml", &test)
	if err != nil {
		log.Fatalf("%v", err)
	}
}
```

File *./test.toml*:

```toml
a.b = "hello"
a = 7
```

This unexpectedly passes.

```ShellSession
$ ./main 
$ echo $?
0
```

Note that `a` as a direct ancestor of `a.b` was redefined as a number, but the library did not detect this.

When we add an extra record in *./test.toml*:

```toml
a.b = "hello"
a.c = "hello"
a = 7
```

the library detects the error.

```ShellSession
$ ./main 
2025/08/24 15:38:10 toml: line 3 (last key "a"): Key 'a' has already been defined.
$ echo $?
1
```

Why did I distinguish between the single and the double record cases: I found this comment in *parse.go*:

```go
		// Normally redefining keys isn't allowed, but the key could have been
		// defined implicitly and it's allowed to be redefined concretely. (See
		// the `valid/implicit-and-explicit-after.toml` in toml-test)
		//
		// But we have to make sure to stop marking it as an implicit. (So that
		// another redefinition provokes an error.)
```

Note that in *valid/implicit-and-explicit-after.toml* there is no direct ancestor's re-definition, which is comletely ok:

```toml
[a.b.c]
answer = 42

[a]
better = 43
```

PR #486, opened 2026-06-09
reject redefining an implicit table as a non-table value

sahvx655-wq:
While tracing how dotted keys register their parent tables I hit a document the parser should reject but accepts: `a.b = "hello"` then `a = 7`, and likewise `[tbl]` with `a.b = 1` then `a = 2`. `setValue` reaches its implicit-redefinition branch, clears the implicit marker and returns early without comparing the incoming value against the table it shadows, so the conflicting scalar assignment is swallowed. Adding a second dotted key before the scalar already raised an error, which is what pointed me at that early return; left as is the file disagrees with the spec and two consumers can end up with different views of `a` (issue #451).

The early return is only correct when the redefinition keeps `a` a table, as in the valid implicit-then-explicit `[a]` case where the value is a freshly built map. It now returns early only when the new value is itself a `map[string]any`, and otherwise raises the existing "already been defined" error. Keeping the check inside `setValue` means dotted keys and table headers both get the same handling without each caller needing its own guard.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -256,6 +256,23 @@ func TestDecodeErrors(t *testing.T) {
 	}
 }
 
+// An inline table inside an array marks its array key implicit, and an empty
+// key inside that table records its type against the parent key. Together that
+// made the implicit-redefinition check above reject the array as if it were a
+// scalar overwriting a table. It's a valid array of tables, not a redefinition.
+func TestDecodeImplicitArrayNotRedefined(t *testing.T) {
+	for _, in := range []string{
+		`l = [{"" = 0}]`,
+		`l = [{"" = 0}, 1]`,
+		`l = [{"" = {"" = 0}}, 1]`,
+	} {
+		var v any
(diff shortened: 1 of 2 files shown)


sahvx655-wq:
Pushed a fix for the CIFuzz failure, which my last change had introduced. The round-trip target tripped on input like `l = [{"" = 0}]`: an inline table inside an array marks the array key implicit, and the empty key inside that table makes setType record its type against the parent key, so isArray stopped recognising `l` as an array. setValue then fell into the new implicit-redefinition branch and rejected a valid array of tables as if a scalar were overwriting a table.

Sorted it by gating the check on the existing value rather than the incoming one: only reject when a table is being overwritten by a non-table, since an array of tables is stored as []map[string]any and isn't a redefinition. Added a decode test for the array-with-empty-key case, the original error cases still hold, and the round trip the fuzzer flagged passes again.

sahvx655-wq:
gentle ping

PR #491, opened 2026-06-28
decode: allow merging map values into existing data

nelsam:
I have a use case where I'm unmarshaling several files in a `conf.d` directory. For each file I unmarshal, I want values in the file to overwrite values from previous files, but leave all of the other data as is. That works already for structs, but when unmarshaling to a map it would lose data. As far as I could tell, it's just maps - types that implement Unmarshaler work as I expected, structs work as I expected, slices get replaced but that also seems correct.

Since this change *could* potentially break existing uses, I've hidden the behavior behind `decoder.Merge(true)`. If you don't tell the decoder to merge data, it works the same as before.

---

I've added a test, but some rough repro steps:

1. Create a map of "default" data:
```go
dflt := map[string]any{
  "foo": map[string]any{
    "bar": "baz",
  },
}
```
2. Unmarshal a file to the map from [1]:
```go
// someFile:
// [foo]
//   bacon: "eggs"
toml.NewDecoder(someFile).Decode(&dflt)
```
3. Check `dflt["foo"]["bar"]`

what I expected: `"baz"`

what I saw: `nil`

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -120,12 +120,19 @@ const (
 type Decoder struct {
 	r       io.Reader
 	maxNest int
+	merge   bool
 }
 
 // NewDecoder creates a new Decoder.
 func NewDecoder(r io.Reader) *Decoder {
 	return &Decoder{r: r, maxNest: 128}
 }
+
+// Merge sets whether dec merges new data with old data in values it decodes to.
+// When true, dec will merge the existing struct fields and map values rather
+// than replacing them. Slices (lists) will still be replaced - implement
+// Unmarshaller on a custom slice type to merge slices.
+func (dec *Decoder) Merge(merge bool) { dec.merge = merge }
 
 // MaxTableNesting sets the maximum table nesting to prevent using an
 // inordinate amount of system resources on deeply nested tables.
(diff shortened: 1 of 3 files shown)


decode.go
```diff
@@ -356,7 +380,7 @@ func (md *MetaData) unifyMap(mapping any, rv reflect.Value) error {
 
 	tmap, ok := mapping.(map[string]any)
 	if !ok {
-		if tmap == nil {
+		if mapping == nil {
```
nelsam, line 383: I made this change because `tmap == nil` is *always* true. It looks like the intent was to check if `mapping` was nil and error if it wasn't.

nelsam:
Changing to a draft - if the `Merge` option is allowed, `Primitive` should probably support it too.

nelsam:
I *did* manage to get it to work, but it required much larger changes to the codebase. And then I ran into *new* issues with my use case. Considering how niche my use case is, my gut tells me that the support headache for maintainers might be a larger cost than it's worth.

I'm closing this for now; if anyone else runs into this issue feel free to comment. It's not too hard to implement. There are some oddities around `Primitive` and whatnot, but it wasn't *that* bad.

I'll open a separate issue to discuss my niche use case and whether or not the maintainers would want to support those features. For now, I ended up just writing my own lexer, parser, and decoder. My project's a hobby project anyway, and I really enjoy writing lexing/parsing code.

PR #493, opened 2026-07-04
Speed up Key.String by 21%

trevorprater:
This tightens `Key.String` by doing two small things:

- grow the `strings.Builder` from the key parts instead of the previous `len(k) * 25` heuristic
- scan for TOML bare-key characters by byte, since bare keys are ASCII-only

The public behavior should be unchanged. I also checked the patch with a temporary differential test that compared the new `Key.String` and `maybeQuoted` implementations against the previous rune-scanning implementation over fixed edge cases, empty keys, Unicode, invalid UTF-8 bytes, and 5,000 deterministic randomized keys.

Validation:

```text
go test -count=1 ./...
```

```text
ok  	github.com/BurntSushi/toml	0.210s
?   	github.com/BurntSushi/toml/cmd/toml-test-decoder	[no test files]
?   	github.com/BurntSushi/toml/cmd/toml-test-encoder	[no test files]
?   	github.com/BurntSushi/toml/cmd/tomlv	[no test files]
?   	github.com/BurntSushi/toml/internal	[no test files]
?   	github.com/BurntSushi/toml/internal/tag	[no test files]
?   	github.com/BurntSushi/toml/internal/toml-test	[no test files]
?   	github.com/BurntSushi/toml/ossfuzz	[no test files]
```

Benchmark command:

```text
go test -run '^$' -bench '^BenchmarkKey$' -benchmem -count=20 .
```

`benchstat`:

```text
goos: darwin
goarch: arm64
pkg: github.com/BurntSushi/toml
cpu: Apple M4
       │ baseline-bench.txt │ final-bench.txt │
       │       sec/op       │     sec/op       vs base                │
Key-10           38.72n ± 1%    30.39n ± 2%  -21.52% (p=0.000 n=20)

       │ baseline-bench.txt │ final-bench.txt │
       │        B/op        │      B/op        vs base                │
Key-10           64.00 ± 0%     48.00 ± 0%  -25.00% (p=0.000 n=20)

       │ baseline-bench.txt │ final-bench.txt │
       │     allocs/op      │   allocs/op      vs base                │
Key-10           1.000 ± 0%     1.000 ± 0%  ~ (p=1.000 n=20)
```

I found the candidate while experimenting with Sleepy - https://sleepy.run, then manually reduced the generated diff and validated it locally.

### DIFF
--- a/meta.go
+++ b/meta.go
@@ -95,8 +95,20 @@ type Key []string
 
 func (k Key) String() string {
 	// This is called quite often, so it's a bit funky to make it faster.
+	size := 0
+	if len(k) > 0 {
+		size = len(k) - 1
+	}
+	for _, kk := range k {
+		if kk == "" {
+			size += 2
+		} else {
+			size += len(kk)
+		}
+	}
+
 	var b strings.Builder
-	b.Grow(len(k) * 25)
+	b.Grow(size)
 outer:
 	for i, kk := range k {
 		if i > 0 {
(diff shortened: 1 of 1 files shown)


arp242:
In a more real-world test case:

	% go test -benchmem -bench=BenchmarkExample
	goos: linux
	goarch: amd64
	pkg: github.com/BurntSushi/toml
	cpu: AMD Ryzen 7 PRO 7840U w/ Radeon 780M Graphics

	old:
	BenchmarkExample/decode-16                 21973             59602 ns/op           32762 B/op        450 allocs/op
	BenchmarkExample/encode-16                 29943             38298 ns/op           13026 B/op        427 allocs/op
	BenchmarkDecode/large-doc-16                2889            384526 ns/op          311739 B/op       2740 allocs/op

	new:
	BenchmarkExample/decode-16                 21085             55988 ns/op           30092 B/op        450 allocs/op
	BenchmarkExample/encode-16                 33382             36108 ns/op           12665 B/op        427 allocs/op
	BenchmarkDecode/large-doc-16                2916            386664 ns/op          291171 B/op       2742 allocs/op

That seems ... insignificant. When I added that microbenchmark, there was an overall improvement of ~30%. This is nowhere near that.

TOML will have Unicode bare keys in 1.2 (probably). So then I'll have to revert bits of this as well.

---

> I found the candidate while experimenting with Sleepy, then manually reduced the generated diff and validated it locally.

I think it would be better to only send PRs for actual problems you encounter instead of using AI tools looking for (non-) problems to fix.

PR #505, opened 2026-07-23
fix: return error when encoding with a nil writer

sonnemusk:
## Summary
`NewEncoder(nil)` built a `bufio.Writer` around a nil `io.Writer`, so `Encode` panicked in `Flush`.

Leave the internal buffer nil when the writer is nil, and return a clear error from `Encode`.

## Test plan
- [x] `go test -run TestEncode`
- [x] New unit test for `NewEncoder(nil).Encode(...)`

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -134,7 +134,11 @@ type Encoder struct {
 
 // NewEncoder create a new Encoder.
 func NewEncoder(w io.Writer) *Encoder {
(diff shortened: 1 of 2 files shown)


sonnemusk:
Closing this — it was a defensive nil guard, not something I hit in real use. Callers should not pass nil here; panicking (or failing at the call site) is fine.

PR #506, opened 2026-07-23
fix: bound ErrorWithPosition against out-of-range positions

sonnemusk:
## What

`ParseError.ErrorWithPosition` clamps `Line` / `Col` / `Len` before indexing the input lines.

## Why

Some parse paths produce positions that are zero or past the last line (e.g. EOF-style errors). Formatting those paniced on slice index.

## Test

- `TestErrorWithPositionOutOfRangeDoesNotPanic`

### DIFF
--- a/error.go
+++ b/error.go
@@ -117,18 +117,35 @@ func (pe ParseError) ErrorWithPosition() string {
 		lines = strings.Split(pe.input, "\n")
 		b     = new(strings.Builder)
 	)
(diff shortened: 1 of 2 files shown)


sonnemusk:
Closing to keep my open contributions focused on higher-impact bugs and features. Happy to revisit if a real-world panic/repro shows up for out-of-range positions.

